PR #119, opened 2023-10-24
feat: Chain 188 use openzeppelin owner in deal

AlcibiadesCleinias:
I decided to use our custom kinda diamond owner contract with logic that does not contradict other diamond storage logic. So with this, we have not to mess between __gap[n] allocations Vs diamond storage allocations

Additionally, I renamed Owner.sol to OwnableUpgradableDiamond.sol to represent its similarity to OZ OwnableUpgradable, and to do not make confusion 

Additionally, I show in pr how want to use hardhat-deploy with upgrades for UUPS proxies (I finally found the solution in github issues. we spoke about yesterday)

PS. If this pr is ok, I will add upg to global contract the same way here: https://linear.app/fluence/issue/CHAIN-189/add-missed-upgradability-to-contracts and introduce uses of `ethers.getContract` instead of chains of `contract__factory()` (in this PR presented)
PPS. I want to add upg tests and owner tests as well


contracts/utils/OwnableUpgradableDiamond.sol
```diff
@@ -4,9 +4,14 @@ pragma solidity ^0.8.19;
 
 import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 
-contract Ownable is Initializable {
+/*
+ * @dev Similar (not fully) to the OZ OwnableUpgradable(Initializable, ContextUpgradeable).
+ * @dev The main difference is in Diamond part,
+ * @dev  e.g. from: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2964.
+*/
+contract OwnableUpgradableDiamond is Initializable {
```
elshan-eth, line 12: Я думаю, что это не совсем Diamond паттерн. Я бы не привязывался к нему в названии и в _STORAGE_SLOT. Т.е. кажется что мы просто вынесли storage в отдельный слот и это не совсем Diamond.

AlcibiadesCleinias: okey, I was in doubt when append the suffix:

I was lead by the next arguments:
- generally this ownable in its signature (interface) makes for us something that OZ OwnableUpgradable makes
- especially makes when used as this: `contract ExampleContract is UUPSUpgradeable, OwnableUpgradable {...`
- but the custom ownable contract uses another approach that could be named as diamond, e.g. because of this https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2964 and other related discussions. So, here I thought that this particularity could be emphasized by Diamon keyword

AlcibiadesCleinias: pls, consider my comment. 

I could agree after on without the Diamond keyword as well, because see that one uses of the keyword potentially creates a lot of places to use the word again, e.g. in strorages

elshan-eth: I got it. We can leave the Daimon suffix. Sounds logical.

PR #122, opened 2023-10-26
feat: CHAIN-191 Deploy on IPC subnet

AlcibiadesCleinias:
* add support for EIP1559 for deploy and tasks
* add contract deployments to public subnet
* add subnet and calibration network to config
* add core utils repo: use in hardhat-ethers dependent scripts or in other only ethers dependent scripts


AlcibiadesCleinias:
I add ipc chain support for our ee test

elshan-eth (approved):

PR #128, opened 2023-11-09
feat: add subgraph folder and simple graphQL schemes [CHAIN-199]

AlcibiadesCleinias:
- added integration with hardhat (with todos on how to make json copying more smooth). - add local dev guide
- cover with scheme current figma around deal market
- add todo on mapping inserting

- currently stopped on hardhat CI tasks. Those tasks could give the possibility to create events and check subgraph, find more potential bugs

UPD:
- I added a task to create market offers (npx hardhat createMarket) - now we could check after the task via GUI that API works and how exactly it works
- I left integration in subgraph repo as npm command
- I rewrote scr/configs and src/global a bit, sorry for the fixes in the PR

UPD1:
- add many to many for effectors Vs offers


deploy/0_tokens/0_deploy_tokens.ts
```diff
@@ -1,27 +1,35 @@
 import type { HardhatRuntimeEnvironment } from "hardhat/types";
+import { WAIT_CONFIRMATIONS, DEFAULT_HARDHAT_DEPLOY_SETTINGS } from "../../env";
+import {saveAbiToSubgraph} from "../../utils/exportAbiToSubgraph";
 import { getEIP1559AndHardhatTxArgs } from "../../utils/deploy";
 
 module.exports = async function (hre: HardhatRuntimeEnvironment) {
     const accounts = await hre.getUnnamedAccounts();
     const deployer = accounts[0]!;
+    const TestUSDDeploymentName = 'TestUSD'
+    const FLTDeploymentName = 'FLT'
```
elshan-eth, line 10: I'm not sure it's worth putting this data into constants. When we look at await hre.deployments.deploy, we can immediately see the name and contact name. When we put this into a separate constancy, we cannot instantly match the name and the contact name.

utils/exportAbiToSubgraph.ts
```diff
@@ -0,0 +1,17 @@
+import path from "path";
+import fs from "fs";
+
+const SUBGRAPH_ABI_DIR = path.join(__dirname, "..", "subgraph", "abis");
+
+// TODO: make by 1 hardhat task.
+export function saveAbiToSubgraph(abi: any, contractName: string) {
```
elshan-eth, line 7: Why do we need a separate directory for the subgraph? We have all the artifacts in the “artifacts” directory

elshan-eth: If we still want to put this in a separate directory, then it is better to copy the “artifacts” directory through the command in package.json. Example: npm run update-subgraph-artifacts

subgraph/subgraph.yaml
```diff
@@ -0,0 +1,44 @@
...
+      kind: ethereum/events
+      apiVersion: 0.0.7
+      language: wasm/assemblyscript
+      entities:
+        - Offer
+        - Effector
+        - Peer
+      abis:
+        - name: CoreImpl
+          file: ./abis/CoreImpl.json
```
elshan-eth, line 24: Can we use here "../contracts/artifacts/"?

deploy/0_tokens/0_deploy_tokens.ts
AlcibiadesCleinias: m, the only reason I put it inot constants was: it was written in strings, but I need to reuse the same strings. So, that is why constants come

AlcibiadesCleinias: but your mentioned optimisation could be implemented, I guess.

I am only afraid that in other projects there always were kinda factory-contract {string: string} mapping. Maybe it is outdated behaviour

subgraph/src/mappings.ts
```diff
@@ -0,0 +1,148 @@
...
+    const effectorTuple = event.params.effector
+    const cidToRemove = effectorTuple[0].toString() + effectorTuple[1].toString()
+
+    if (entity.effectors == null) {
+        return
+    }
+
+    // Find effector to remove
+    let effectorsToSave: Array<string> = []
+    for (let i=0; i < entity.effectors!.length; i++) {
```
elshan-eth, line 131: Can we make a different structure so that deletion is done in O(1)?

utils/exportAbiToSubgraph.ts
AlcibiadesCleinias: the better solution exists, here I did not want to stop to implement the best ones, merely was hopping to finish integration that will work

so, better solution is to dive into solution that hardhat-subgraph plugin implement and solve some issue in their repo, and then use it. It is just matter of time. Lets left our solution with TODO mentioned for the current stage, could that work?

deploy/0_tokens/0_deploy_tokens.ts
elshan-eth: Got it. If the name is reused, then this is the better solution. When I first read it, I didn’t immediately notice that they were being reused.

subgraph/subgraph.yaml
AlcibiadesCleinias: no, subgraph needs files with abi's only

utils/exportAbiToSubgraph.ts
elshan-eth: It seems to me that now there is no need to use hardhat-subgraph as we will switch to foundry. Here I mean that it is better if we transfer artifacts through a separate command in package.json, not through js code.

subgraph/src/mappings.ts
AlcibiadesCleinias: subgraph **build** command could not deal with complex structures, I tried to implement tuple (not here), and it was not compilled. I guess I should left TODO here: to check other data structures that could give us better performance

subgraph/subgraph.yaml
elshan-eth: I mean link to the directory with contracts for abi from the artifacts folder

subgraph/src/mappings.ts
elshan-eth: We can store the offer id in  effector as a backlink.

elshan-eth:
```
type Offer @entity {
  """Used in the next figma views:
    - Offer from List Of offers 1.2.
    - Offer 1.2
  """
  id: ID!  # Ref to offerId in contract.
  createdAt: BigInt!
  provider: Bytes!
  pricePerEpoch: BigInt!
  tokenSymbol: String!
  "Currently effector id is merged(prefix + hash) aka 1 CID."
  effectors: [Effector!]
  peers: [Peer!] @derivedFrom(field: "offer")

  "Sum of all connected compute units."
  computeUnitsSum: Int
# deprecated: maxCollateralPerWorker: BigInt
}

"Effector table is obsolete table since it is a simple mapping."
type Effector @entity(immutable: true) {
  id: ID! 
  description: EffectorDescription!
  offer: OfferId
}

type EffectorDescription @entity(immutable: true) {
  "CID"
  id: Bytes!
  Description: String!
}

type Peer @entity {
  "ref to peerId in contract."
  id: ID!
  offer: Offer!
  computeUnits: Int!
}
```

elshan-eth: EffectorDescription - this is a global description for effectors. Those. 1 effector has 1 description.
Effector - This is a table with data linked to the offer. It contains a link to EffectorDescription and Offer.
Offer - has a link to effector

elshan-eth: When we want to delete an effector from an offer, we search for and delete the Effector structure by Id and OfferId.

utils/exportAbiToSubgraph.ts
AlcibiadesCleinias: oh, I get you, mb yes

when I put kinda these commands into deploy ones I follow the same pattern as in the hardhat-subgraph-plugin.

This pattern always insure that you work with updated abis. But I see, that this npm command mb even better to insert into the graph command of build, codegen

subgraph/subgraph.yaml
elshan-eth: Got it, then my comments are not relevant

utils/exportAbiToSubgraph.ts
elshan-eth: I understand what the problem is. Then my comment is not relevant. We can leave it like that for now.

AlcibiadesCleinias: ok, I believe that the best place for this one is subgraph npm command first of all. And hardhat depoy command should not be aware of 'some sort of other package scripts'. I will relocate it

AlcibiadesCleinias: this https://github.com/fluencelabs/deal/pull/128/commits/c14a8bf45d3bb2eee49ac18cec767ed3afe1bcaf
and this: https://github.com/fluencelabs/deal/pull/128/commits/44a4f3c64564ce357bf5ab17651d441d468efa1c

subgraph/src/mappings.ts
AlcibiadesCleinias: so, it is just about many-to-many

AlcibiadesCleinias: ok, it is possible to implement of coz

AlcibiadesCleinias: ok, I resolved this staff. Was too afraid to add many to many via the graph syntax, but it is kinda hacked now

elshan-eth:
The CI didn't work. npm build can't find typechain

AlcibiadesCleinias:
> The CI has not worked. npm build can't find typechain

if you checkout on target branch - the same: it is about scr/clients - we already changed a lot of contracts

AlcibiadesCleinias:
> > The CI has not worked. npm build can't find typechain
> 
> if you checkout on target branch - the same: it is about scr/clients - we already changed a lot of contracts

I can fix it here, I am too afraid about merge conflicts if you are solved it, and will push to market-offer branch

PR #135, opened 2023-11-24
feat: resolve scheme for Deal Screens [chain-202]

AlcibiadesCleinias:
I resolved Deal from Figma I guess, but with several TODO about white/black lists and with unsolved todo about expired status of workers

UPD:
- resolve CHAIN-220
- resolve CHAIN-222
- resolve CHAIN-201

UPD:
- a few events come with new args
- some events renamed


subgraph/subgraph.yaml
```diff
@@ -19,11 +19,15 @@ dataSources:
...
       abis:
         - name: CoreImpl
           file: ./abis/CoreImpl.json
+        - name: DealImpl
+          file: ./abis/DealImpl.json
       eventHandlers:
+        # Offer features.
         - event: MarketOfferRegistered(bytes32,address,uint256,address,(bytes4,bytes32)[])
           handler: handleMarketOfferRegistered
 #          Currently PeerCreated and ComputeUnitCreated are not supported together as nested events.
```
elshan-eth, line 33: Can you test it again? Maybe this error no longer exists

elshan-eth:
createOrLoad{entity} - this one doesn't seem to fit very well
We always have an understanding of when an entity needs to be created and when to load it. In this case, it may happen that we forgot to write down the entity correctly somewhere, but we will not receive an error about this.

For example: 
`export function handleWithdrawn(event: Withdrawn): void {
    let deal = createOrLoadDeal(event.address.toHex())
    deal.depositedSum = deal.depositedSum.minus(event.params.amount)
    deal.save()
}`
We definitely know that here we need to load the object. 
Better solution:
`export function handleWithdrawn(event: Withdrawn): void {
   let deal = Deal.load(event.address.toHex())!;
    deal.depositedSum = deal.depositedSum.minus(event.params.amount)
    deal.save()
}`

subgraph/src/contracts.ts
```diff
@@ -1,9 +1,22 @@
...
+} from '../generated/Core/DealImpl'
 import {Address, Bytes} from "@graphprotocol/graph-ts";
 
+
+export function getOfferInfo(contractAddress: Address, offerId: string): CoreImpl__getOfferResultValue0Struct {
+    const contract = CoreImpl.bind(contractAddress)
+    return contract.getOffer(Bytes.fromHexString(offerId))
+}
+
 export function getComputeUnit(contractAddress: Address, unitId: Bytes): CoreImpl__getComputeUnitResultValue0Struct {
```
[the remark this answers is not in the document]
elshan-eth: If it's because of problem with reading multiple events, we need to figure out why this is so

subgraph/src/mappings/deal.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function handleDeposited(event: Deposited): void {
+    let deal = createOrLoadDeal(event.address.toHex())
+    deal.depositedSum = event.params.amount.plus(deal.depositedSum)
+    deal.save()
+}
+
+export function handleWithdrawn(event: Withdrawn): void {
+    let deal = createOrLoadDeal(event.address.toHex())
+    deal.depositedSum = deal.depositedSum.minus(event.params.amount)
```
elshan-eth, line 19: deal.withdrawalSum

subgraph/src/mappings/core.ts
```diff
@@ -0,0 +1,216 @@
...
+        '[handleDealCreated] New deal created: {} by: {}',
+        [event.params.owner.toString(), dealAddress.toString()]
+    )
+
+    let deal = createOrLoadDeal(dealAddress.toHex())
+    deal.createdAt = event.block.timestamp
+    deal.client = event.params.owner
+
+    // Fetch other data from the contract.
+    const contract = getDealContract(dealAddress)
```
elshan-eth, line 194: This shouldn't have been a call. We must take everything from events.

subgraph/src/mappings/deal.ts
```diff
@@ -0,0 +1,51 @@
...
+    deal.appCID = getEffectorCID(cid)
+    deal.save()
+}
+
+// Joined to Deal.
+export function handleComputeUnitJoined(event: ComputeUnitJoined): void {
+    let deal = createOrLoadDeal(event.address.toHex())
+
+    let computeUnit = createOrLoadComputeUnit(event.params.unitId.toHex())
+    computeUnit.addedToDeal = deal.id
```
elshan-eth, line 42: `deal` is more good naming then `addedToDeal`

subgraph/schema.graphql
```diff
@@ -32,24 +43,68 @@ E.g. to use
...
+    id: ID! # Set to `deal.id.concat(effector.id)`
+    deal: Deal!
+    effector: Effector!
+}
+
+type Deal @entity {
+  id: ID!  # Ref to offerId in contract.
+  appCID: String!
+  createdAt: BigInt!
+  client: Bytes!
```
elshan-eth, line 90: client -> owner

AlcibiadesCleinias:
> createOrLoad{entity} - this one doesn't seem to fit very well
> 
> We always have an understanding of when an entity needs to be created and when to load it. In this case, it may happen that we forgot to write down the entity correctly somewhere, but we will not receive an error about this.
> 
> For example:
> 
> ```ts
> export function handleWithdrawn(event: Withdrawn): void {
>     let deal = createOrLoadDeal(event.address.toHex())
>     deal.depositedSum = deal.depositedSum.minus(event.params.amount)
>     deal.save()
> }
> ```
> 
> We definitely know that we need to load the object here.
> 
> Better solution:
> 
> ```ts
> export function handleWithdrawn(event: Withdrawn): void {
>    let deal = Deal.load(event.address.toHex())!;
>     deal.depositedSum = deal.depositedSum.minus(event.params.amount)
>     deal.save()
> }
> ```

so, here I basically tried to follow the same scheme from the graph developers, like here: 
- https://github.com/graphprotocol/graph-network-analytics-subgraph/blob/master/src/mappings/controller.ts
- https://github.com/graphprotocol/graph-network-analytics-subgraph/blob/master/src/mappings/curation.ts
- somewhere they really used simple load, e.g. https://github.com/graphprotocol/graph-network-analytics-subgraph/blame/master/src/mappings/curation.ts#L109 (and this behaviour is kinda new one [8 month old], and not used too much across the repo)

Thus, I show you kinda code style I oriented on.
What I think from perspective of dev:

when you use createOrLoad:
- you simplify logic for yourself: by accessing the model with 1 method instead of choosing, 
- you always may reimplement id mechanics and so one, and you use for it 1 module instead of prepping and etc.
- if we relay on order of event: and thus for different events we write somewhere create, somewhere load, if event order changes - you will need to skim the subgraph code and improve the order.
- I am kindly not sure (because of my dimmed overview of the subgraph and its distributed polling mechanics) that each of the node load events in the same order and it may not be changed in the future.
- based on above point: even if the handler of the 2nd event via createOrLoad creates an entity that frontend should not use (because the entity will look be with some zero necessary fields [bc it is only seconds event]) - here we have classic kinda bug-feature of the eventual consistency. and that is okey, as for email systems: when you see merely wrong counter of received email, but after 1-2 sec everything comes on the right place

subgraph/src/contracts.ts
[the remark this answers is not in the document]
AlcibiadesCleinias: I tried to simplify, and with the time I could implement reading from the event

subgraph/schema.graphql
AlcibiadesCleinias: and in the figma he is client, but okey, lets be consistent with contracts instead

subgraph/subgraph.yaml
AlcibiadesCleinias: I got the same error:

```json
nceManager
subgraph-graph-node-1  | Nov 27 11:25:43.745 INFO Syncing 1 blocks from Ethereum, code: BlockIngestionStatus, blocks_needed: 1, blocks_behind: 1, latest_block_head: 280, current_block_head: 279, provider: localhost-rpc-0, component: EthereumPollingBlockIngestor
subgraph-graph-node-1  | Nov 27 11:25:43.806 INFO Done processing trigger, gas_used: 6395359327, data_source: Core, handler: handleComputeUnitAddedToDeal, total_ms: 115, transaction: 0x62bd…ce50, address: 0x789e…940b, signature: ComputeUnitAddedToDeal(bytes32,address), sgd: 1, subgraph_id: QmTEPDvawVoSrkaQsK7YHzXz2wQnsJrmbY2jmP8repLMUq, component: SubgraphInstanceManager
subgraph-graph-node-1  | Nov 27 11:25:43.900 INFO Done processing trigger, gas_used: 1293849270, data_source: Deal, handler: handleComputeUnitJoined, total_ms: 93, transaction: 0x62bd…ce50, address: 0x4f81…4fe1, signature: ComputeUnitJoined(bytes32), sgd: 1, subgraph_id: QmTEPDvawVoSrkaQsK7YHzXz2wQnsJrmbY2jmP8repLMUq, component: SubgraphInstanceManager
subgraph-graph-node-1  | Nov 27 11:25:44.055 INFO Done processing trigger, gas_used: 6395359327, data_source: Core, handler: handleComputeUnitAddedToDeal, total_ms: 154, transaction: 0x62bd…ce50, address: 0x789e…940b, signature: ComputeUnitAddedToDeal(bytes32,address), sgd: 1, subgraph_id: QmTEPDvawVoSrkaQsK7YHzXz2wQnsJrmbY2jmP8repLMUq, component: SubgraphInstanceManager
subgraph-postgres-1    | 2023-11-27 11:25:47.700 UTC [46] LOG:  could not receive data from client: Connection reset by peer
subgraph-postgres-1    | 2023-11-27 11:25:47.700 UTC [46] LOG:  unexpected EOF on client connection with an open transaction
subgraph-graph-node-1 exited with code 137


```

AlcibiadesCleinias: okey, I then up container again and it works, I have even checked via GUI.

I guess, the error occurs when while it is still indexing I try to access the GUI

AlcibiadesCleinias: pff, lets close it, I rm todo

AlcibiadesCleinias: okey, I fix everything according to the new event hadnler, and the constant error occurs, I left it in the branch: https://github.com/fluencelabs/deal/tree/error/137docker

AlcibiadesCleinias: in the PR I will not commit the uncommented code

subgraph/src/contracts.ts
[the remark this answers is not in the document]
AlcibiadesCleinias: so, I read the event already: grep: `export function handleComputeUnitCreated(event: ComputeUnitCreated): void {`

I use this getter for the ComputeUnitAddedToDeal where I need access the info about peerId of CU

subgraph/subgraph.yaml
elshan-eth: https://github.com/graphprotocol/graph-node/issues/2325

subgraph/src/mappings/core.ts
AlcibiadesCleinias: no contract calls

AlcibiadesCleinias:
I also found a bug of myself: https://github.com/fluencelabs/deal/pull/135/commits/f4f1d5c6bba70d503ba525e7ee80c7b633117803

subgraph/src/mappings/deal.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export function handleAppCIDChanged(event: AppCIDChanged): void {
+    let deal = Deal.load(event.address.toHex()) as Deal
+    const cid = changetype<AppCID>(event.params.newAppCID)
+    deal.appCID = getEffectorCID(cid)
+    deal.save()
+}
+
+// Joined to Deal.
+export function handleComputeUnitJoined(event: ComputeUnitJoined): void {
```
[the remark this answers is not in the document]
AlcibiadesCleinias: rename event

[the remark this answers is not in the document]
AlcibiadesCleinias: https://github.com/fluencelabs/deal/pull/135/commits/a4ae93f47443a9176f8c11cb7028c02efa7fc968

PR #163, opened 2023-12-12
feat: paginator in graphql [chain-238]

AlcibiadesCleinias:


ts-client/src/dealExplorerClient/indexerClient/indexerClient.ts
```diff
@@ -45,6 +74,21 @@ export class IndexerClient {
     return await this.offersClient.OfferQuery(variables);
   }
 
+  // TODO: create generic instead.
```
AlcibiadesCleinias, line 77: yeah, I did few tries, but no decided to economy time

elshan-eth:
You can merge everything related to the subgraph without review if you don't mind. It seems that now there is not much time to review ts-client. Thanks!

ts-client/src/dealExplorerClient/indexerClient/indexerClient.ts
```diff
@@ -29,6 +30,20 @@ export class IndexerClient {
     this.providersClient = getProvidersSdk(client);
   }
 
+  async getTotalProviders(variables?: ProvidersQueryQueryVariables, offset: number = 0): Promise<number> {
```
elshan-eth, line 33: Does this solution run in O(N)?

elshan-eth: There are two ways to solve this:
- change the design
- let the frontend request data by page. I mean frontend can make N requests to get pages. Each query will return the result gradually. The frontend will be able to render on fontend page by page by increasing the count on the frontend side.

elshan-eth: O(n) solutions are terrible. We shouldn't use them. Please be attentive to them. I understand that TheGraph does not have the ability to calculate count. But then you need to come up with a solution that will be better than waiting in O(n).

AlcibiadesCleinias: in frontend we have had paginator already. So the task is set.

It is not O(n)
it is light weigh request (only ids requested) with batch of 1000 (current, max 6k - not tested).

AlcibiadesCleinias: @elshan-eth , checkout this.PAGINATOR_ENITIES_LIMIT - it limits to 1000 ids per batch

elshan-eth: Yes, but for 10,000 elements, that would be 10 queries, right?

elshan-eth: As I see it, recursion is used here. This method is run with a batch of 1000 elements for all elements. If so, then it's O(n).

elshan-eth: For N * 1000 entities, we will have N queries.

AlcibiadesCleinias: I will manage a call

Currently the impl as it is because we have subgraph and design.

Mb we will suggest a minor change in design, because around paginator there is another good big trouble exist

elshan-eth: We can make no one async call to get all totals. This could be a dynamic loading that the frontend will receive from different requests and not just one.

elshan-eth: Change design is ok too. But we definitely need to avoid such decisions as the current one. This is very important to think about before we start developing this.

AlcibiadesCleinias: dynamic background pagination processing -  that is something that might be called as complex frontend, that we were not agreed to create

elshan-eth: Dynamic background pagination processing doesn't sound difficult. 

I mean here that we shouldn’t use crutches if we encounter a problem. But we need to think about how to do it differently. Currently, the method will have O(n) execution time. If one query takes 300 ms at best, then 10 queries (for 10,000 elements) will take 3 seconds, and 100 queries 30 seconds.

PR #171, opened 2023-12-19
feat: add deploy script with basic auth for stage

AlcibiadesCleinias:


subgraph/fluence-graph.sh
```diff
@@ -0,0 +1,35 @@
...
+# Validate envs.
+if [ "$BASIC_AUTH_FLUENCE_STAGE" = "" ]; then
+  echo "Please provide 'BASIC_AUTH_FLUENCE_STAGE' env."
+  exit 0
+fi
+
+# Choose auth string.
+basic_auth=""
+if [ "$BASIC_AUTH_FLUENCE_STAGE" != "" ]; then
+  basic_auth=${BASIC_AUTH_FLUENCE_STAGE}@
```
nahsi, line 22:
```suggestion
  basic_auth="${BASIC_AUTH_FLUENCE_STAGE}@"
```

subgraph/fluence-graph.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Validate envs.
+if [ "$BASIC_AUTH_FLUENCE_STAGE" = "" ]; then
+  echo "Please provide 'BASIC_AUTH_FLUENCE_STAGE' env."
+  exit 0
+fi
+
+# Choose auth string.
+basic_auth=""
+if [ "$BASIC_AUTH_FLUENCE_STAGE" != "" ]; then
```
nahsi, line 21:
```suggestion
if [[ -z $BASIC_AUTH_FLUENCE_STAGE ]]; then
```

`[[` is bash builtin command, `[` is legacy command outside of bash, if I'm not mistaken.

It is a minor thing but according to modern bash guidelines it is better to use `[[` :)
http://mywiki.wooledge.org/BashGuide/TestsAndConditionals

`-z` check means `var is undefined or empty`

subgraph/fluence-graph.sh
```diff
@@ -0,0 +1,35 @@
...
+# >> export BASIC_AUTH_FLUENCE_STAGE=<check .example.env> && ./fluence-graph.sh
+
+# Args:
+# $1 - action {deploy, create}.
+
+# Envs:
+# - BASIC_AUTH_FLUENCE_STAGE
+
+# Validate envs.
+if [ "$BASIC_AUTH_FLUENCE_STAGE" = "" ]; then
```
nahsi, line 14:
```suggestion
if [[ -z $BASIC_AUTH_FLUENCE_STAGE ]]; then
```

subgraph/fluence-graph.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
```
nahsi, line 1:
```suggestion
#! /bin/env bash
```

subgraph/fluence-graph.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Choose auth string.
+basic_auth=""
+if [ "$BASIC_AUTH_FLUENCE_STAGE" != "" ]; then
+  basic_auth=${BASIC_AUTH_FLUENCE_STAGE}@
+fi
+
+SUBGRAPH_NAME=fluence-deal-contracts
+
+if [ "$1" = "deploy" ]; then
```
nahsi, line 27:
```suggestion
if [[ $1 = "deploy" ]]; then
```

subgraph/fluence-graph.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Deploy subgraph on Fluence for stage env with subgraph name: $SUBGRAPH_NAME..."
+  graph deploy --node https://${basic_auth}graph-node-admin.fluence.dev/ --ipfs https://${basic_auth}graph-node-ipfs.fluence.dev --network stage --network-file config/networks.json --version-label 0.0.0 ${SUBGRAPH_NAME}
+elif [ "$1" = "create" ]; then
```
nahsi, line 30:
```suggestion
elif [[ "$1" = "create" ]]; then
```

subgraph/fluence-graph.sh
nahsi, line 27: I personaly would use `case` here.
```
case $1 in
  deploy)
    echo ... ;;
  create)
    echo ... ;;
  *)
     echo error
     exit 1
  ;;
esac
```

But it is a matter of taste

nahsi (approved):

subgraph/fluence-graph.sh
AlcibiadesCleinias: even for strings?

PR #175, opened 2023-12-21
feat: matching [chain-205]

AlcibiadesCleinias:
- I restructured clients dir according to context of Explorer client and (new) Matcher client
- add new client with its indexer
- add tiny aka `config` solution for indexer URLs config
- move the logic of matching search into indexer
- refactor subgraph a bit
- resolve minWorkers check (it reverts with custom error) (& integrate into ts-client)
- I skip tests for ts-client/client/... - coz it comes from capacity directly
- [x] ts-client fix filter boundaries
- [x] optional: tests on matcher client and methods

Discuss (that out of scope of this PR)
- `// TODO: Catch CU does not exist error and pass.` - implement exact error
- // TODO: check peer in capacity commitment
- // TODO: check max peers per provider (workers?)
- // TODO: event and support in the subgraph.


script/CreateMarket.s.sol
```diff
@@ -90,7 +90,7 @@ contract CreateMarket is Depoyments, Script {
 
         address[] memory createdDeals = new address[](dealCount);
         for (uint32 i = 0; i < dealCount; i++) {
-            uint256 newMinWorkers = startMinWorkers + i;
+            uint256 newMinWorkers = 1 + i;
```
elshan-eth, line 93: Is it legacy or you don't need startMinWorkers?

src/core/modules/market/Matcher.sol
```diff
@@ -51,58 +50,60 @@ contract Matcher is Offer, IMatcher {
...
 
         CIDV1 memory appCID = deal.appCID();
         uint256 creationBlock = deal.creationBlock();
 
         bool isDealMatched = false;
 
-        LinkedListWithUniqueKeys.Bytes32List storage offerList = _getOffersList();
-
-        bytes32 currentOfferId = offerList.first();
+        for (uint i = 0; i < computeUnits.length; ++i) {
```
elshan-eth, line 63: I think we need to provide several arrays `bytes32[] offers and bytes32[][] computeUnits`. Maybe peers too but I don't  sure. If we walk through the units, we can double-read and compare offer params.

src/core/modules/market/Matcher.sol
```diff
@@ -51,58 +50,60 @@ contract Matcher is Offer, IMatcher {
...
 
-            currentOfferId = offerList.next(currentOfferId);
+            if (freeWorkerSlotsCurrent == 0) {
+                // TODO: possible feature of signalling to user that matched fully.
+                break;
+            }
         }
 
+        uint minWorkers = deal.minWorkers();
+        if (minWorkers > dealComputeUnitCount + freeWorkerSlots - freeWorkerSlotsCurrent) revert minWorkersNotMatchedError(minWorkers);
```
elshan-eth, line 105: Check with @evgenyponomarev

src/core/modules/market/Offer.sol
```diff
@@ -154,10 +149,10 @@ contract Offer is BaseModule, IOffer {
 
         computePeer.unitCount--;
 
-        offerStorage.freeComputeUnitByPeerId[peerId].remove(unitId);
         offerStorage.computeUnitIdsByPeerId[peerId].remove(unitId);
 
         delete offerStorage.computeUnits[unitId];
+        // TODO: event and support in the subgraph.
```
elshan-eth, line 155: Can you add an event in this PR?

src/core/modules/market/interfaces/IMatcher.sol
```diff
@@ -6,6 +6,8 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "src/deal/interfaces/IDeal.sol";
 
 interface IMatcher {
+    // ------------------ Errors ------------------
+    error minWorkersNotMatchedError(uint _minWorkers);
```
elshan-eth, line 10: The naming convention of the errors is the same as those of the events. minWorkersNotMatchedError -> MinWorkersNotMatched.

ts-client/src/dealMatcherClient/dealMatcherClient.ts
```diff
@@ -0,0 +1,236 @@
...
+      deal.pricePerWorkerEpoch as string,
+      deal.effectors.map((effector) => {
+        return effector.effector.id;
+      }),
+      deal.paymentToken.id,
+      targetWorkerToMath,
+      minWorkersToMatch,
+      deal.maxWorkersPerProvider,
+    );
+  }
```
shamsartem, line 235: Small improvements that I can suggest
```suggestion
  /**
   * Get compute units to match provided DealId (address).
   *
   * 1. Fetches the deal and its configuration from the Indexer backend
   * 2. Scraps compute units (page by page) until one of the conditions:
   * - the end of matched offers/peers/compute units
   * - all target compute units found.
   */
  async getMatchedOffersByDealId(
    dealId: string,
  ): Promise<GetMatchedOffersResult> {
    const { deal } = await this._indexerClient.getDeal({ id: dealId });

    if (!deal) {
      throw new DealNotFoundError(dealId);
    }

    const alreadyMatchedComputeUnits = deal.addedComputeUnits?.length ?? 0;
    const targetWorkerToMath = deal.targetWorkers - alreadyMatchedComputeUnits;
    const minWorkersToMatch = Math.max(
      deal.minWorkers - alreadyMatchedComputeUnits,
      0,
    );

    if (!deal.effectors) {
      throw new Error(`deal ${dealId} has no effectors`);
    }

    return this.getMatchedOffers(
      deal.pricePerWorkerEpoch as string,
      deal.effectors.map((effector) => {
        return effector.effector.id;
      }),
      deal.paymentToken.id,
      targetWorkerToMath,
      minWorkersToMatch,
      deal.maxWorkersPerProvider,
    );
  }
```

ts-client/src/dealMatcherClient/dealMatcherClient.ts
```diff
@@ -0,0 +1,236 @@
...
+   * - all target compute units found.
+   */
+  async getMatchedOffersByDeal(
+    dealId: string,
+  ): Promise<GetMatchedOffersResult> {
+    const data: DealQueryQuery = await this._indexerClient.getDeal({
+      id: dealId,
+    });
+    if (!data.deal) {
+      throw new DealNotFoundError(dealId);
```
shamsartem, line 209: If I am understanding correctly if this error happens we will see just the dealId without any propper error message. Or am I wrong? Would be good to see a readable error mesasage

ts-client/src/dealMatcherClient/dealMatcherClient.ts
```diff
@@ -0,0 +1,236 @@
...
+    let minWorkersToMatch = minWorkers - alreadyMatchedComputeUnits;
+    if (minWorkersToMatch < 0) {
+      minWorkersToMatch = 0;
+    }
+
+    if (!deal.effectors) {
+      throw new Error("Assert: deal: " + dealId + " has no effectors.");
+    }
+    return await this.getMatchedOffers(
+      deal.pricePerWorkerEpoch as string,
```
shamsartem, line 226: I think probably this can be fixed in the place where it is defined so it is string there already, instead of making a type assert here

ts-client/src/dealMatcherClient/dealMatcherClient.ts
```diff
@@ -0,0 +1,236 @@
...
+    const targetWorkerToMath = dealTargetWorkers - alreadyMatchedComputeUnits;
+    let minWorkersToMatch = minWorkers - alreadyMatchedComputeUnits;
+    if (minWorkersToMatch < 0) {
+      minWorkersToMatch = 0;
+    }
+
+    if (!deal.effectors) {
+      throw new Error("Assert: deal: " + dealId + " has no effectors.");
+    }
+    return await this.getMatchedOffers(
```
shamsartem, line 225: I would suggest to refactor this function so it accepts a js objec with these values instead of accepting so many args. Would certainly be hard to work with this later cause you waould have to carefully think about the arguments order and make sure not to mess up anything

ts-client/src/dealMatcherClient/dealMatcherClient.ts
```diff
@@ -0,0 +1,236 @@
...
+    console.info("pricePerWorkerEpoch = " + pricePerWorkerEpoch);
+    console.info("effectors = " + JSON.stringify(effectors));
+    console.info("paymentToken = " + paymentToken);
+    console.info("targetWorkerSlotToMatch = " + targetWorkerSlotToMatch);
+    console.info("minWorkersToMatch = " + minWorkersToMatch);
+
+    // TODO.
+    console.warn(
+      `maxWorkersPerProvider = ${maxWorkersPerProvider} currently ignored.`,
+    );
```
shamsartem, line 74: I think this should be removed cause it can log unnecessary stuff
If you need something like that for debuggin purposes there is a debug lib which we use for everything, so you can use that and when you need these logs you can set an env variable in order to see them (or in browser I think you would need to set a value in the local storage)
Also it would be easier to log if you have a js object as one arg to this function - then you can log just one thing `JSON.stringify(arg, null, 2)` to see the whole thing together

script/CreateMarket.s.sol
AlcibiadesCleinias: was: market were created with deals config: minWorkers: n, targetWorkers: n

For me this market was useless, I left the pure example on how to create diversity (b/w deals) and difference b/w minWorkers/targetWorkers in one Deal

src/core/modules/market/Offer.sol
AlcibiadesCleinias: If I add event in this PR, I will need to add a handler in the subgraph, so too many features for the PR.

Thus, I left TODO because not sure if you changed it in your PR, but I do not want to miss it.
let me leave it out of scope: https://linear.app/fluence/issue/CHAIN-258/todo-event-and-support-in-the-subgraph

src/core/modules/market/interfaces/IMatcher.sol
AlcibiadesCleinias: оке, он май вей

ts-client/src/dealMatcherClient/dealMatcherClient.ts
AlcibiadesCleinias: oke, and + await in the end

shamsartem: await in the end does nothing in this case, it's already a promise. It's ok for async function to return a promise directly

ts-client/src/dealMatcherClient/dealMatcherClient.ts
AlcibiadesCleinias: oke, I will add a proper message as well. 

But I believe that the message of this error should not be propagated to the client. 
On client side, there should be a catch of `DealNotFoundError` and frontend-user-compatible message after

shamsartem: Yeah if frontend needs a propper message, sure, they will handle it in any case. But at least it should be obvious what happened. Unfortunately js doesn't have some agreed-upon result types to aprropriatly handle errors. You can learn the function throws at all only when you look at the source code or encounter the error yourself

ts-client/src/dealMatcherClient/dealMatcherClient.ts
AlcibiadesCleinias: for now, I will leave it as it is, sorry, need to migrate to another client. I left a todo

ts-client/src/dealMatcherClient/dealMatcherClient.ts
AlcibiadesCleinias: I agree, so I will implement an interface for that

ts-client/src/dealMatcherClient/dealMatcherClient.ts
AlcibiadesCleinias: But I believe that logs for out and in transactions should be.

In general - when send: info, when receive: debug.
To control log flow - there should be param. Like LOGGING_INFO: WARNING.

Seems, you you do not need this log, you could disable it with env. Seems for log.info you could do it

src/core/modules/market/Matcher.sol
AlcibiadesCleinias: resolved

src/core/modules/market/Matcher.sol
AlcibiadesCleinias: oke, I resolved this finally

PR #180, opened 2023-12-28
chore: Fix CI after deal refactoring

nahsi:


docker/docker-compose.yml
```diff
@@ -9,27 +9,33 @@
...
     build:
       context: ../
-      dockerfile: ./docker/Dockerfile.anvil
+      dockerfile: ./docker/Dockerfile
+      target: chain-rpc
     ports:
       - "8545:8545"
 
-  deploy-script:
+  chain-deploy-script:
```
AlcibiadesCleinias, line 21: but to be honest it is about compilation of artifacts and deploy contracts

subgraph/fluence-graph.sh
```diff
@@ -1,35 +1,75 @@
...
+      echo "Unknown argument '$1'."
+      help
+      exit 1
+  esac
+done
+
+# Prepare variables depending on network
+case "$network" in
+  local)
+    GRAPHNODE_URL="${GRAPHNODE_URL:-http://localhost:8020}"
```
AlcibiadesCleinias, line 48: GRAHP_NODE, but that is okey too

docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+ apt-get install -y --no-install-recommends \
+   nodejs && \
+ apt-get clean && \
+ rm -rf \
+	/tmp/* \
+	/var/lib/apt/lists/* \
+	/var/tmp/*
+
+COPY ./deployments /app/deploymens
+RUN cd subgraph && npm install && npm run compile && npm run import-config-networks
```
AlcibiadesCleinias, line 76: mb it is better to reuse commands from make - like so: 
https://github.com/fluencelabs/deal/tree/capacity?tab=readme-ov-file#develop-with-deal-infrastructure

`make build-npms`?

docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+RUN make build-contracts
+
+FROM contracts as chain-rpc
+ENTRYPOINT ["make"]
+CMD ["start-local-chain"]
+
+FROM contracts as chain-deploy-script
+ENTRYPOINT ["make"]
+CMD ["depoy-docker"]
```
AlcibiadesCleinias, line 47: wow, this image deploys docker inside?)

docker/Dockerfile
```diff
@@ -0,0 +1,82 @@
+FROM ubuntu:22.04 as foundry
```
AlcibiadesCleinias, line 1: please, leave a note what this dockerfile is about, and how to use it. Now we have a kinda mess with Dockerfile in docker/ folder =(

Maybe even it could simplify this flow? https://github.com/fluencelabs/deal/tree/capacity?tab=readme-ov-file#develop-with-deal-infrastructure

docker/docker-compose.yml
nahsi: I try to use the same name everywhere - in CI, in releases etc. Since this script deploys contracts to chain I called it like this.

docker/Dockerfile
nahsi: I don't want to build deal-ts-clients since it will increase already bloated image size.

docker/Dockerfile
nahsi: Yep it just calls make target.

docker/Dockerfile
nahsi: It is better to use multistaged Dockerfile if images share something (in this case building contracts). This way it will reuse cached images instead of building new every time.

It also speeds up CI from ~15 minutes to ~5.

I will add comments but it doesn't matter much since docker-compose will work the same way as before.

AlcibiadesCleinias: I meant comments in Readme md, may be event improve the flow in the last section where we demonstrate how to create local dev env

subgraph/fluence-graph.sh
nahsi: Artem already uses this env vars in his setup. Let's leave it like this?

AlcibiadesCleinias (approved):
pre approve

AlcibiadesCleinias:
we also need to leave simple docker-compose with chain dependencies, not with postgres and etc - that needed for subgprah and subpograh itself. We also need in the same task to deprecate subpgrah/docker-compose.yml or take into scope as well

nahsi:
> we also need to leave simple docker-compose with chain dependencies, not with postgres and etc - that needed for subgprah and subpograh itself. We also need in the same task to deprecate subpgrah/docker-compose.yml or take into scope as well
> 
> lets left as separate task, pls

You can start only needed services with `docker-compose up -d foo bar`

nahsi:
> we have no instruction on how to use images from CI to recreate the infra with
> 
> -node,
> 
> * deployed contracts and
> * subgprah
> 
> lets left it as separate task, mention me as well pls

I don't understand what you mean. CI images are used only in CI. For local development you can use docker-compose file to build images locally

deployments/local.json
```diff
@@ -1,61 +1,61 @@
...
-    "codeHash": "0x7b9e077766ddb539fa294262e14fb58a44394c5d8adff5a0a876498662cd7ae7",
-    "creationCodeHash": "0x6a52dc0dfe68860b5b01338d3c25187f3f57f9440a95610ce5fc903f498c1abd"
+    "blockNumber": 2,
+    "codeHash": "0x2a5d64b37bb4e92b0f53bc559f4486ffaa3341b629ac0155655fc1c13540c2a1",
+    "creationCodeHash": "0x8270f8fac67ecc5b84ef8cf82c24a622af763ace09e83866ed59f6fed2320a12"
   },
   "Multicall3": {
     "addr": "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0",
-    "blockNumber": 0,
+    "blockNumber": 2,
```
AlcibiadesCleinias, line 46: yeah, definitely we need to rm local files from git scope

AlcibiadesCleinias:
> > we have no instruction on how to use images from CI to recreate the infra with
> > -node,
> > 
> > * deployed contracts and
> > * subgprah
> > 
> > lets left it as separate task, mention me as well pls
> 
> I don't understand what you mean. CI images are used only in CI. For local development you can use docker-compose file to build images locally

what if someone want to get contract infra for development os something? I believe it will be easier for him to merely use prepared containers rather that tinker with forge, foundry, npm, etc

Also this documented flow could reveal a bit flow from CI. Improve understanding of the flow from fCLI as well

nahsi:
> > > we have no instruction on how to use images from CI to recreate the infra with
> > > -node,
> > > 
> > > * deployed contracts and
> > > * subgprah
> > > 
> > > lets left it as separate task, mention me as well pls
> > 
> > 
> > I don't understand what you mean. CI images are used only in CI. For local development you can use docker-compose file to build images locally
> 
> what if someone want to get contract infra for development os something? I believe it will be easier for him to merely use prepared containers rather that tinker with forge, foundry, npm, etc
> 
> Also this documented flow could reveal a bit flow from CI. Improve understanding of the flow from fCLI as well

Then they will use `fluence local up`

AlcibiadesCleinias:
> > > > we have no instruction on how to use images from CI to recreate the infra with
> > > > -node,
> > > > 
> > > > * deployed contracts and
> > > > * subgprah
> > > > 
> > > > lets left it as separate task, mention me as well pls
> > > 
> > > 
> > > I don't understand what you mean. CI images are used only in CI. For local development you can use docker-compose file to build images locally
> > 
> > 
> > what if someone want to get contract infra for development os something? I believe it will be easier for him to merely use prepared containers rather that tinker with forge, foundry, npm, etc
> > Also this documented flow could reveal a bit flow from CI. Improve understanding of the flow from fCLI as well
> 
> Then they will use `fluence local up`

and what about e.g. frontend from network explorer project?)

PR #184, opened 2024-01-04
fix: [chain-268]

AlcibiadesCleinias:
~~* also fixes for deploy script~~

~~- [ ] same fix for network client~~

upd: I realised that fixes should come from capacity branch (https://github.com/fluencelabs/deal/pull/185)
coz need to deliver to Artem them [fixes]

upd; in this PR:

- build-ts-clients make command
- fix deployment - add inits modules for contracts
- [x] integrate e2e test with CC
- [x] update readmies in notion and scheme?
- [ ] implement check for CC in offchain matcher client: ref to https://github.com/fluencelabs/deal/pull/186/files#diff-97d19024a8961b684ebd08d3f933ae3afe208b64788d5005b4b9bd9dc96b290fR97
- [ ] optional: `// TODO: if needToRedeployMarket or needToRedeployCapacity - impl proxy should be update with impl.`


ts-client/tests/unittests/client.test.ts
```diff
@@ -1,34 +1,38 @@
...
-// TODO: fix tests and unskip.
-test.skip("test env", async () => {
-  expect(await getDeployment("local")).toEqual({
-    core: "0x09635F643e140090A9A8Dcd712eD6285858ceBef",
-    flt: "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512",
-    usdc: "0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0",
-    chainId: 31337,
-  });
+test("#getDeployment", async () => {
+  // TODO: not sure if have to support local env in git!
```
folex, line 7: why not?

Makefile
```diff
@@ -20,16 +20,25 @@ build-contracts: ## Build contracts
 	@forge build
 	@echo "\033[0;32mSuccess! Build of contracts completed.\033[0m"
 
-build-npms: ## Build npms
+build-ts-clients: ## Build ts-clients
```
elshan-eth, line 23: `build-ts-clients`, maybe `build-ts-client` or Do we have several clients?

AlcibiadesCleinias: there are several clients already

AlcibiadesCleinias: dealClient - but it is more DealContractClient, ContractApi

ExplorerClient
DealMatcherClient

AlcibiadesCleinias: also I want to rename ts-client to ts-clients though

AlcibiadesCleinias: but okey, I see. Until we rename ts-client to ts-clients, we should not pass here s

ts-client/tests/unittests/client.test.ts
AlcibiadesCleinias: from deploy to deploy it might be changed on local. coz it is local. Thus, basically we should not save local artifacts at all into the git

elshan-eth (approved):

PR #190, opened 2024-01-12
feat!: add whitelist [fixes CHAIN-273 CHAIN-274 CHAIN-275]

elshan-eth:
# Changelog

## DealFactory: 
 - function `deployDeal` has two additional arguments: 
`IConfig.AccessType providersAccessType_` - Access level for a deal  (None, whitelist, blacklist)
`address[] calldata providersAccessList_` - List of addresses in this access list

## Global whitelist 
- add a global whitelist that allows you to limit providers that can create CC
- add two events:
` event WhitelistAccessGranted(address account);`
` event WhitelistAccessRevoked(address account);`

## Deal 
- add a access list and access type (whitelist, blacklist)
- new events: 
` event ProvidersAccessTypeChanged(AccessType newAccessType);`
`event ProviderAddedToAccessList(address provider);`
`event ProviderRemovedFromAccessList(address provider);`


elshan-eth:
Need to fix ts-client tests

AlcibiadesCleinias:
> Need to fix ts-client tests

`npm test unittests` are resolved successfully 

`npm test integrational` - required created and running infra (currently those tests are not in the CI)

src/core/modules/market/Matcher.sol
```diff
@@ -61,25 +61,35 @@ abstract contract Matcher is Offer, IMatcher {
...
 
         // Go through offers.
         for (uint256 i = 0; i < offers.length; ++i) {
             bytes32 offerId = offers[i];
             Offer memory offer = getOffer(offerId);
+
             if (
-                pricePerWorkerEpoch < offer.minPricePerWorkerEpoch || paymentToken != offer.paymentToken
-                    || !_hasOfferEffectors(offerId, effectors)
+                !deal.hasProviderAccess(offer.provider) || pricePerWorkerEpoch < offer.minPricePerWorkerEpoch
```
AlcibiadesCleinias, line 79: deal has provider access - it is really difficult to understand what is happening with these namings

src/deal/Config.sol
```diff
@@ -71,7 +77,9 @@ contract Config is OwnableUpgradableDiamond, IConfig {
         uint256 targetWorkers_,
         uint256 maxWorkersPerProvider_,
         uint256 pricePerWorkerEpoch_,
-        CIDV1[] calldata effectors_
+        CIDV1[] calldata effectors_,
+        AccessType providersAccessType_,
```
AlcibiadesCleinias, line 81: why do we have only 1 access type per deal?

AlcibiadesCleinias: I thought we need black list and white list all together

src/core/modules/capacity/Whitelist.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    function __Whitelist_init_unchained(bool isWhitelistEnabled) internal onlyInitializing {
+        _getWhitelistStorage().isWhitelistEnabled = isWhitelistEnabled;
+    }
+
+    // #endregion
+
+    // #region ----------------- Public View -----------------
+    function hasAccess(address account) external view returns (bool) {
```
AlcibiadesCleinias, line 41: naming is so strange.
mb something:
- `approved`
- `isApproved`
- 'whitelisted'
- 'isWhitelisted'


Btw, For me it seems that WhiteList contract is an abstract only for whitelisting, and if we want to disable those whitelist features in a product we should not do it in the WhiteList itself. We should do it in another abstraction on a layer higher, imho

src/core/modules/capacity/Whitelist.sol
```diff
@@ -0,0 +1,60 @@
...
+        WhitelistStorage storage s = _getWhitelistStorage();
+        return !s.isWhitelistEnabled || s.isWhitelisted[account];
+    }
+    // #endregion
+
+    // #region ----------------- Public Mutable -----------------
+    function addToWhitelist(address account) external onlyOwner {
```
AlcibiadesCleinias, line 48: mb simple `add`?

src/core/modules/capacity/Whitelist.sol
```diff
@@ -0,0 +1,60 @@
...
+        _getWhitelistStorage().isWhitelisted[account] = true;
+
+        emit AddedToWhitelist(account);
+    }
+
+    function removeFromWhitelist(address account) external onlyOwner {
```
AlcibiadesCleinias, line 54: mb simple remove?

src/core/modules/market/Matcher.sol
```diff
@@ -94,29 +104,27 @@ abstract contract Matcher is Offer, IMatcher {
...
                 emit ComputeUnitMatched(peerId, computeUnitId, creationBlock, appCID);
+
+                computeUnitCountInDealByProvider++;
                 freeWorkerSlotsCurrent--;
 
                 if (!isDealMatched) {
                     isDealMatched = true;
                 }
-
-                if (freeWorkerSlotsCurrent == 0) {
```
AlcibiadesCleinias, line 116: why did you remove this break?

seems, it is used to stop iteration when we riched target deal worker?

src/core/modules/market/Offer.sol
```diff
@@ -200,7 +200,8 @@ abstract contract Offer is BaseModule, IOffer {
         offerStorage.computeUnitIdsByPeerId[peerId].remove(unitId);
 
         delete offerStorage.computeUnits[unitId];
-        // TODO: event and support in the subgraph.
+
+        emit ComputeUnitRemoved(peerId, unitId);
```
AlcibiadesCleinias, line 204: I added task for this: https://linear.app/fluence/issue/CHAIN-277/support-cu-removed-in-subgraph-client

AlcibiadesCleinias (commented):
About integrational test: we both solve maxWorkersPerProvider TODO in contract in different branches.

That is okey, MB better to merge firstly your PR than I could update my branch and resolve integration test, coz anyway I have it to resolve rather in my branch with my updates or in this, with another updates on maxWorkersPerProvider

src/deal/Config.sol
elshan-eth: A whitelist will always be more important than a blacklist. That is, they are not together.

src/core/modules/capacity/Whitelist.sol
elshan-eth: I think `hasAccess` is a goode name but `approved` and `isApproved` is better. I will rename it.


I don’t quite understand why this needs to be done at a higher level

src/core/modules/market/Matcher.sol
elshan-eth: I moved this rule above

test/core/Matcher.t.sol
```diff
@@ -201,6 +202,27 @@ contract DealMock {
...
+
+    function getComputeUnitCount() external view returns (uint256) {
+        return computeUnitCount;
+    }
+
+    function getComputeUnitCount(address provider) external view returns (uint256) {
+        return computeUnitCountByProvider[provider];
+    }
+
+    function maxWorkersPerProvider() external pure returns (uint256) {
```
AlcibiadesCleinias, line 225: why so, why not to save maxWorkerPerProvider into mock?

AlcibiadesCleinias: and simply return through public variable

elshan-eth: I didn't quite get your question

AlcibiadesCleinias: instead of providing method with pure getMax, MB better to leave public variable: https://github.com/fluencelabs/deal/pull/192/files#diff-d75734009f15a4478d68d1a1981ce994c7380b0fa517148aa244488c9fd25df3R177

AlcibiadesCleinias:
> Need to fix ts-client tests

not sure, but mb those fixes could help you: 

- https://github.com/fluencelabs/deal/pull/192/files#diff-904435590130207c656dbea3291d8ce0001307ed4103e804760a97d6d807b434R102
- and this link: https://github.com/fluencelabs/deal/pull/192/files#diff-cc3765daa57de0f2ce6e5a88c6f305afdfcfdf29a891c8ef6df22f27a2385583L107 that leads to implementing maxWorkersPerProvider check on the part of the matcher ts-client

but again, MB better to skip the integrational test in this branch since I fixed it on the part of the feature of this branch: `maxWorkersPerProvider`, and on other part: `deal whitelist` feature I could support on request

test/core/Matcher.t.sol
elshan-eth: I think that both ways are normal. I don't see any advantages or disadvantages between them.

src/core/modules/market/Matcher.sol
elshan-eth: I change to `isProviderAllowed`

AlcibiadesCleinias (approved):

AlcibiadesCleinias:
@elshan-eth, let's write a changelog with breaking changes (events, signature of methods) directly in PR and link this changelog in on-chain channel by mentioning: Folex, Masha, Alexey Pr.. Thus, they will be notified that they need to update: miro board, aqua typescript flow, etc.

e.g. for this PR:

# Changelog
- new events in Deal:
```
    event ProvidersAccessTypeChanged(AccessType newAccessType);
    event ProviderAddedToAccessList(address provider);
    event ProviderRemovedFromAccessList(address provider);
```

PR #194, opened 2024-01-16
fix: chain 279 cosmetic and more

AlcibiadesCleinias:
# Breaking Change

- ~new event: `UpdateProviderApprove` with method: `function updateProviderApprove(address provider, bool changeApproveTo) external;` (for more check method descriptions)~
- I refactored types.ts in dealExplorerClient/ -> it goes into types/schemes.ts and types/filters.ts

# Change
also it closes: 
- CHAIN-297 (more filtration based on approved provider and if offer status "is active")
- CHAIN-305 (approved for provider by core owner)


src/core/modules/market/Offer.sol
```diff
@@ -101,11 +101,20 @@ abstract contract Offer is BaseModule, IOffer {
     function setProviderInfo(string calldata name, CIDV1 calldata metadata) external {
         require(bytes(name).length > 0, "Name should be not empty");
 
-        _getOfferStorage().providers[msg.sender] = ProviderInfo({name: name, metadata: metadata});
+        _getOfferStorage().providers[msg.sender] = ProviderInfo({name: name, metadata: metadata, approved: false});
 
         emit ProviderInfoUpdated(msg.sender, name, metadata);
     }
 
+    function updateProviderApprove(address provider, bool changeApproveTo) onlyCoreOwner external {
```
elshan-eth, line 109: we have Whitelist contract for it

grantAccess -> event WhitelistAccessGranted

src/core/modules/market/interfaces/IOffer.sol
```diff
@@ -10,9 +10,11 @@ import "src/deal/base/Types.sol";
 /// @dev Offer contract is responsible for managing the offers in the market
 interface IOffer {
     // ------------------ Types ------------------
+    /// @param approved If provider address approved by Fluence (Core Owner).
     struct ProviderInfo {
         string name;
         CIDV1 metadata;
+        bool approved;
```
elshan-eth, line 17: we have method isApproved in Whitelist contract

AlcibiadesCleinias: but whitelist is a feature for deal

AlcibiadesCleinias: alright, whitelist in capacity contract should be used instead

PR #203, opened 2024-01-17
feat(tests): Introduce integration tests package

akim-bow:
In this task i tried to setup dedicated package for integration tests.
The idea is to put all integration scenarios in that folder

Also made some overall quality of life impovements


integration-tests/src/happy.test.ts
```diff
@@ -10,9 +10,13 @@
...
 // TODO: more variates not only 1to1 match.
-import { describe, test, expect } from "vitest";
+import { describe, test, expect, assert } from "vitest";
 
-import { ContractsENV, DealClient, DealMatcherClient } from "../../src";
+import {
+  type ContractsENV,
+  DealClient,
+  DealMatcherClient,
+} from "@fluencelabs/deal-ts-clients";
```
AlcibiadesCleinias, line 19: how to run those tests from local on changes in this client?

subgraph/package-lock.json
```diff
@@ -758,17 +758,6 @@
         "url": "https://github.com/prettier/prettier?sponsor=1"
       }
     },
```
AlcibiadesCleinias, line 760: why do we need those changes in pr?

AlcibiadesCleinias (commented):
with those renaming to happy test I can not understand how I will improve test scenarios for dealMatcherFlow in the future.

Since I am not seeing context (e.g. other tests) I am thinking that it is wrong to rename this e2e test for matching flow only into integration-test/happy.test.ts

Also, I can not understand how to run this test locally after e.g. I update dealMatcherClient and want to test before updates push to remote

integration-tests/src/happy.test.ts
akim-bow: You can `npm pack` ts-client package and put the created archive to package.json as a resolution.
The current resolution url goes to our private registry and it could be easily replaced with path to archive.
It's not perfect as it requires you to explicitly pack the client package. But it could be improved later if we change `npm` to `pnpm` which solves these cases without the need for packing.

subgraph/package-lock.json
akim-bow: Seems like this lockfile change came after i made some fixes in repo structure and configs.
Basically, the removed entry have been provided `prettier` package resolution in case Node.JS version smaller than 14. Probably someone ran `npm i` with very old Node.JS version.

akim-bow:
> with those renaming to happy test I can not understand how I will improve test scenarios for dealMatcherFlow in the future.
> 
> Since I am not seeing context (e.g. other tests) I am thinking that it is wrong to rename this e2e test for matching flow only into integration-test/happy.test.ts

I will redo renaming if you see previous naming fit. Probably i have mistakenly considered the test as a general one and happy path for CC creation.

integration-tests/src/happy.test.ts
AlcibiadesCleinias: so, the flow: I want to make some dev on matcherClient, I want to test them after, and to do so I need to pack, rewrite line in tests: `} from "@fluencelabs/deal-ts-clients";` on my local path, if everything is okey I need to revert `} from "my/local/path";` and push commit finally?

akim-bow: I wrote README, please check does it answer your question

AlcibiadesCleinias: if flow is described we could close it

AlcibiadesCleinias (approved):
with 1 comment

integration-tests/src/deal-matcher.test.ts
```diff
@@ -22,16 +22,16 @@ import { ethers } from "ethers";
 
 // TODO: from env.
 const TEST_NETWORK: ContractsENV = "local";
-const TEST_RPC_URL = "http://localhost:8545";
-const DEFAULT_SUBGRAPH_TIME_INDEXING = 10000;
+const TEST_RPC_URL = "http://akim-dev.dev.fluence.dev:8545";
```
AlcibiadesCleinias, line 25: best: implement fetch from env mechanics and with localhost default one.
minimal: revert to localhost (;

integration-tests/src/deal-matcher.test.ts
```diff
@@ -268,7 +258,10 @@ describe("#getMatchedOffersByDealId", () => {
 
       console.log("---- Deal Matching ----");
       console.info(`Find matched offers for dealId: ${dealId}...`);
-      const dealMatcherClient = new DealMatcherClient(TEST_NETWORK);
+      const dealMatcherClient = new DealMatcherClient(
+        TEST_NETWORK,
```
AlcibiadesCleinias, line 264: same

PR #224, opened 2024-01-25
fix!: Mv from erc20 to native

elshan-eth:
- remove fluenceToken like ERC20. fluenceToken is native token.


AlcibiadesCleinias (dismissed):
could you mark that you remove fluenceToken in the PR changelog please

also, could you support those changes in ts-client/tests/integrational-tests?

PR #233, opened 2024-02-03
chore: Add IPC to docker

nahsi:


script/CreateMarket.s.sol
```diff
@@ -0,0 +1,237 @@
+// SPDX-License-Identifier: Apache-2.0
```
AlcibiadesCleinias, line 1: it was deprecated several ages ago, lets rm it

PR #237, opened 2024-02-06
chore: core contract unit tests and minor conditions fix

crystalbit:
- added Core.t.sol for Core.sol test coverage
- changed require condition in Core's `initializeModules` a bit

### DIFF
--- a/test/core/Core.t.sol
+++ b/test/core/Core.t.sol
@@ -2,8 +2,6 @@
 pragma solidity ^0.8.19;
 
 import {Test, console2} from "forge-std/Test.sol";
-// import "forge-std/console.sol";
-// import "forge-std/Vm.sol";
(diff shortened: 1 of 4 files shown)


test/core/Core.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        deployment = DeployDealSystem.deployDealSystem();
+        bool initialized = deployment.initialized;
+
+        if (initialized) {
+            return;
+        }
+    }
+
+    function test_CoreHasInitializedValues() external {
+        assertFalse(address(deployment.core.dealImpl()) == address(0), "Deal impl not initialized in Core");
```
elshan-eth, line 32: I find this difficult to read. It might be better to replace it with assertEq. Maybe I'm wrong, what do you think about this?

PR #247, opened 2024-02-09
feat: min and maxProtocolVersion

crystalbit:
Added protocol versions. Deal now has `protocolVersion`, and Offer has `minProtocolVersion` and `maxProtocolVersion`.
Also there are global `minProtocolVersion` and `maxProtocolVersion` being set by DAO/multisig

# Breaking
- added `uint256 protocolVersion` to `DealProxy` initializer
- added `uint256 protocolVersion` to `DealCreated` event
- added `uint256 minProtocolVersion` and `uint256 maxProtocolVersion` parameters to `registerMarketOffer` function
- added `uint256 minProtocolVersion` and `uint256 maxProtocolVersion` to `MarketOfferRegistered` event

# Change Log
- Matcher is now aware of protocolVersion, min and max protocol versions


elshan-eth (approved):

AlcibiadesCleinias (approved):

src/core/interfaces/IGlobalConst.sol
```diff
@@ -35,6 +37,14 @@ interface IGlobalConst {
     /// @return min rematching epoches for all deals
     function minDealRematchingEpoches() external view returns (uint256);
 
+    /// @dev Returns min protocol version
```
AlcibiadesCleinias, line 40: but why we need this version? mb we could add description?)

AlcibiadesCleinias: it seems we will have further discussion even. So basically we want to have end-to-end versioning across Fluence resources and we do not want do not want to support matching offers with deals with incompatible versions

mb this should be written here somehow

AlcibiadesCleinias (approved):

PR #263, opened 2024-02-15
feat: indexer for cli - get offer with some data [chain-356]

AlcibiadesCleinias:
- add contribution flow
- add example client use based on DealExplorerClient and Matcher style
- resolve fetch of offer with its effectors from subgraph


shamsartem (approved):
nice. thanks

PR #268, opened 2024-02-19
move dealfactory out of market

crystalbit:


AlcibiadesCleinias (changes requested):
test are missed, we need to support those changes in subgraph. It is not difficult, maybe you want to ask me tomorrow?

Also, we need to write ChangeLog in this PR to link other teammates to introduction of those breaking changes

AlcibiadesCleinias (changes requested):
Also, I noticed that this new feature is not supported in our ts-client/client/... where we get the interfaces of contract (market) in `cli` and contract addresses: https://github.com/fluencelabs/deal/blob/main/ts-client/src/client/client.ts#L48

this client is our kinda main interface client for `cli` to interact with our deal contracts. Thus, this changes in signature of main **market** contract should be supported by us in this PR

crystalbit:
> tests are failed on subpraph module, we need to support those changes in subgraph. It is not difficult, maybe you want to ask me tomorrow?
> 
> Also, we need to write ChangeLog in this PR to link other teammates to introduction of those breaking changes

ok

AlcibiadesCleinias (approved):
sgtm

just do not forget to redeploy contracts and write breaking changes for cli
and refer on this changelog in our channels

PR #271, opened 2024-02-19
feat: add LOCAL_CHAIN_BLOCK_MINING_INTERVAL and note useful envs

AlcibiadesCleinias:
* add LOCAL_CHAIN_BLOCK_MINING_INTERVAL
* more envs to show


.example.env
```diff
@@ -0,0 +1,14 @@
+# Private key to use for sender of transactions (deploy, etc) in forge scripts for {testnet, kras, stage}.
```
folex, line 1: Will these vars be configurable via env in Docker container?

.example.env
```diff
@@ -0,0 +1,14 @@
...
+PRIVATE_KEY=0x0
+# Wait confirmation in scripts and tests.
+WAIT_CONFIRMATIONS=5
+# Local Chain (Anvil Node).
+LOCAL_CHAIN_BLOCK_MINING_INTERVAL=15
```
folex, line 6: Let's set to 1 by default maybe? Who needs 15 seconds on local chain?

.example.env
AlcibiadesCleinias: it relates to deploy scripts. If envs set in docker and then makefile commands run in docker, values from env will be used

.example.env
AlcibiadesCleinias: we have right now 15 everywhere, need to check first of all the setup for 1 sec... I am too afraid and do not want to debug it right now in the case

.example.env
folex: https://github.com/fluencelabs/deal/blob/main/docker/Dockerfile#L48-L51

Seems like it does.

folex (approved):

.example.env
AlcibiadesCleinias: so, does that mean that that is okie and as expected?

folex: Seems that it will be available to configure via envs to docker container, yes.

.example.env
folex: Okay

PR #304, opened 2024-02-26
fix!: add some fixes

elshan-eth:
- Fix vesting. Add vestingDuration and vestingCount in global vars.
- Add depositAmount to deployDeal 
- fix dealExit for CU
- Change error from string to error type (error CapacityCommitmentIsNotActive,CapacityCommitmentIsActive)
- fix deal reward


src/core/modules/market/interfaces/IDealFactory.sol
```diff
@@ -57,6 +57,7 @@ interface IDealFactory {
     function deployDeal(
         CIDV1 calldata appCID_,
         IERC20 paymentToken_,
+        uint256 depositAmount_,
```
AlcibiadesCleinias, line 60: Could we please add a readable description of this to the interface?

src/core/modules/capacity/Capacity.sol
```diff
@@ -314,7 +316,9 @@ contract Capacity is UUPSUpgradeable, MulticallUpgradeable, CapacityConst, White
         uint256 expiredEpoch = _expiredEpoch(cc);
         CCStatus status = _commitCommitmentSnapshot(cc, peer, currentEpoch_ - 1, expiredEpoch);
 
-        require(status == CCStatus.Inactive || status == CCStatus.Failed, "Capacity commitment is active");
+        if (status != CCStatus.Inactive && status != CCStatus.Failed) {
+            revert CapacityCommitmentIsActive(status);
```
AlcibiadesCleinias, line 320: ohh, with this custom errors our local logs will be fkced coz Anvil does not support custom errors.

Do we really ready to go like this? mb string is better?

elshan-eth: Yes, we need it for nox

PR #309, opened 2024-03-01
feat: Introduce integration tests

akim-bow:


.github/docker-compose.yml
```diff
@@ -0,0 +1,95 @@
+# yaml-language-server: $schema=schemas/docker-compose.json
```
AlcibiadesCleinias, line 1: why we do not use merely docker/docker-compose.yml?

@nahsi

integration-tests/src/commitment.test.ts
```diff
@@ -0,0 +1,145 @@
...
+describe(
+  "Capacity commitment",
+  () => {
+    beforeAll(async () => {
+      provider = new ethers.JsonRpcProvider(TEST_RPC_URL);
+      signer = await provider.getSigner();
+      contractsClient = new DealClient(signer, TEST_NETWORK);
+    });
+
+    test.only("CC can be removed before deposit", async () => {
```
AlcibiadesCleinias, line 37: only detected!

integration-tests/src/confirmations.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+  type ICapacity,
+  type IDeal,
+  type IDealFactory,
+  type IMarket,
+} from "@fluencelabs/deal-ts-clients";
+import type { TypedEventLog } from "@fluencelabs/deal-ts-clients/dist/typechain-types/common.js";
+import { type ContractTransactionResponse } from "ethers";
+
+export async function confirmEvents<
```
AlcibiadesCleinias, line 11: Am I understanding right that what you want: is to get event value from a transaction? (if yes, I knew with help of Elshan) about this method: https://github.com/fluencelabs/deal/blob/main/ts-client/tests/integration/scripts/create-pure-market.test.ts#L394

also, naming of confirmations.ts is smth that is not understandable. mb module should be named as `events.ts` and method is `checkLastEvents` | `checkLastNEvents` ?

integration-tests/src/deal.test.ts
```diff
@@ -0,0 +1,177 @@
...
+  ENDED,
+}
+
+describe(
+  "Deal tests",
+  () => {
+    beforeAll(async () => {
+      provider = new ethers.JsonRpcProvider(TEST_RPC_URL);
+      // signer = new ethers.Wallet(
+      //   "0x2AE9D3ABFA37CEA42FE829298CAC07CC25AF213A4C82BD74CBEFAD305C749B4F",
```
AlcibiadesCleinias, line 31: private key

integration-tests/src/deal.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        deal.filters.Deposited,
+        1,
+        depositTx,
+      );
+      expect(dealDepositEvent?.args.amount).toEqual(amountToDeposit);
+
+      freeBalance = await deal.getFreeBalance();
+      console.log("deal free balance", freeBalance);
+
+      // console.log("Deal withdraw...");
```
AlcibiadesCleinias, line 149: comments

AlcibiadesCleinias: ok, mb commented is okie.. so, TODO?

integration-tests/src/helpers.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const createdCommitments = peerIds.length;
+
+  // Fetch created commitmentIds from chain.
+  const filterCreatedCC = capacity.filters.CommitmentCreated;
+  const capacityCommitmentCreatedEvents = await capacity.queryFilter(
+    filterCreatedCC,
+    fromBlock,
+  );
+  const capacityCommitmentCreatedEventsLast = capacityCommitmentCreatedEvents
```
AlcibiadesCleinias, line 81: check out for this: https://github.com/fluencelabs/deal/blob/main/ts-client/tests/integration/scripts/create-pure-market.test.ts#L394

.github/docker-compose.yml
akim-bow: I do use `docker/docker-compose.yml` now

integration-tests/src/deal.test.ts
akim-bow: Added TODO

integration-tests/src/commitment.test.ts
akim-bow: Do you mean detected CC? What does that mean?

integration-tests/src/helpers.ts
akim-bow: Not sure what you're trying to say but it's simply bad code. I will improve it later. For now i want to add a single test suit `deal.test.ts` as fast as possible

integration-tests/src/confirmations.ts
akim-bow: Ideally i want to retrieve logs with events directly from transaction, yeah. But your code doesn't really suits me because it isn't type safe. Types are lost because you're using low level functions. I will figure out something later, thanks for the suggestion.

Regarding function name - i'm ok with the renaming it to `checkEvents` if you see fit. The function doesn't check for the last events now.

.github/docker-compose.yml
AlcibiadesCleinias: so, do we need this docker-compose though?

integration-tests/src/helpers.ts
akim-bow: I mean my code, which needs to be improved :)

AlcibiadesCleinias (approved):

PR #326, opened 2024-03-07
fix!: deal statuses

elshan-eth:
- Add new SMALL_BALANCE status. It's when the balance is less than `targetWorkers * 2 epoches * price`.

### DIFF
--- a/src/core/modules/market/Matcher.sol
+++ b/src/core/modules/market/Matcher.sol
@@ -52,8 +52,7 @@ abstract contract Matcher is Offer, IMatcher {
         MatcherStorage storage matcherStorage = _getMatcherStorage();
 
         IDeal.Status dealStatus = deal.getStatus(); 
-        require(dealStatus != IDeal.Status.ENDED, "Matcher: deal is ended");
-        require(dealStatus != IDeal.Status.INSUFFICIENT_FUNDS, "Matcher: deal has insufficient funds");
+        require(dealStatus == IDeal.Status.ACTIVE || dealStatus == IDeal.Status.NOT_ENOUGH_WORKERS, "Matcher: deal is not active");
 
         uint256 lastMatchedEpoch = matcherStorage.lastMatchedEpoch[address(deal)];
         uint256 currentEpoch = core.currentEpoch();
--- a/src/deal/Deal.sol
+++ b/src/deal/Deal.sol
@@ -90,7 +90,7 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
         uint256 prevEpoch = _globalCore().currentEpoch() - 1;
 
         DealStorage storage dealStorage = _getDealStorage();
-        dealStorage.maxPaidEpoch = prevEpoch;
+        dealStorage.maxPaidEpoch = 0;
         dealStorage.lastCommitedEpoch = prevEpoch;
     }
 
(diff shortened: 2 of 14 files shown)


src/deal/Deal.sol
```diff
@@ -90,7 +90,7 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
         uint256 prevEpoch = _globalCore().currentEpoch() - 1;
 
         DealStorage storage dealStorage = _getDealStorage();
-        dealStorage.maxPaidEpoch = prevEpoch;
+        dealStorage.maxPaidEpoch = 0;
```
folex, line 93: Let's leave a comment on why that makes sense?

src/deal/interfaces/IDeal.sol
```diff
@@ -13,7 +13,8 @@ interface IDeal is IConfig, IWorkerManager {
         INSUFFICIENT_FUNDS,
         ACTIVE,
         ENDED,
-        NOT_ENOUGH_WORKERS
+        NOT_ENOUGH_WORKERS,
+        SMALL_BALANCE
```
folex, line 17: Let's comment what this means and how SMALL_BALANCE is different from INSUFFICIENT_FUNDS?

src/core/modules/market/Matcher.sol
```diff
@@ -52,8 +52,7 @@ abstract contract Matcher is Offer, IMatcher {
         MatcherStorage storage matcherStorage = _getMatcherStorage();
 
         IDeal.Status dealStatus = deal.getStatus(); 
-        require(dealStatus != IDeal.Status.ENDED, "Matcher: deal is ended");
-        require(dealStatus != IDeal.Status.INSUFFICIENT_FUNDS, "Matcher: deal has insufficient funds");
+        require(dealStatus == IDeal.Status.ACTIVE || dealStatus == IDeal.Status.NOT_ENOUGH_WORKERS, "Matcher: deal is not active");
```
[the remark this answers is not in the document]
elshan-eth: We need to match the deal if it is ACTIVE or NOT_ENOUGH_WORKERS. With other statuses, we don't need to match.

src/deal/Deal.sol
```diff
@@ -164,16 +164,22 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
...
+
+        if (maxPaidEpoch != 0 && _globalCore().currentEpoch() > dealStorage.maxPaidEpoch) {
             return Status.INSUFFICIENT_FUNDS;
+        } else if (freeBalance < _globalCore().minDealDepositedEpoches() * pricePerWorkerEpoch() * targetWorkers()) {
+            return Status.SMALL_BALANCE;
+        } else if (currentWorkerCount < minWorkers()) {
+            return Status.NOT_ENOUGH_WORKERS;
         } else {
             return Status.ACTIVE;
         }
```
folex, line 179: If status is INSUFFICIENT_FUNDS, you would calculate freeBalance and currentWorkerCount for nothing 🤔

Maybe do
```suggestion
        uint256 maxPaidEpoch = dealStorage.maxPaidEpoch;
        if (maxPaidEpoch != 0 && _globalCore().currentEpoch() > dealStorage.maxPaidEpoch) {
            return Status.INSUFFICIENT_FUNDS;
        } 
        
        uint256 freeBalance = getFreeBalance();
        if (freeBalance < _globalCore().minDealDepositedEpoches() * pricePerWorkerEpoch() * targetWorkers()) {
            return Status.SMALL_BALANCE;
        } 
        
        uint256 currentWorkerCount = getWorkerCount();
        if (currentWorkerCount < minWorkers()) {
            return Status.NOT_ENOUGH_WORKERS;
        }
       
        return Status.ACTIVE;
```

src/core/modules/market/Matcher.sol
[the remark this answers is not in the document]
folex: ah, I misread the condition. My bad.

folex (approved):

PR #334, opened 2024-03-09
fix: deal preCommit overflow

elshan-eth:


src/deal/Deal.sol
```diff
@@ -130,7 +130,7 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
 
     function _postCommitPeriod(
         DealStorageUtils.Balance memory balance,
-        uint256 prevEpoch,
+        uint256 commitEpoch,
```
folex, line 133: should other `_postCommitPeriod` call sites be changed as well?

src/deal/Deal.sol
```diff
@@ -96,32 +96,32 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
     // ------------------ Privat Functions ------------------
     function _preCommitPeriod(
         DealStorageUtils.Balance memory balance,
-        uint256 prevEpoch,
+        uint256 commitEpoch,
```
folex, line 99: should other `_preCommitPeriod` call sites be changed as well? There are many of them 🤔

PR #340, opened 2024-03-12
feat(integration-tests): Provider cleanup

akim-bow:


integration-tests/src/commitment.test.ts
```diff
@@ -20,6 +20,21 @@ let provider: JsonRpcProvider;
 let signer: JsonRpcSigner;
 let contractsClient: DealClient;
 
+async function sendProof(capacityContract: ICapacity, coreContract: ICore, cuId: string, proofs: number, epoches: number) {
```
folex, line 23: You don't use `proofs` argument

akim-bow: `while (sentProofRounds++ < proofs) {`

integration-tests/src/commitment.test.ts
```diff
@@ -19,6 +34,57 @@ const DEFAULT_TEST_TIMEOUT = 180000;
...
+) {
+  const difficulty = await capacityContract.difficulty();
+  let sentProofRounds = 0;
+  let currentEpoch = await coreContract.currentEpoch();
+  console.log(currentEpoch, "currentEpoch");
+  const provider = capacityContract.runner?.provider;
+  assert(provider, "no provider");
+  let currentNonce = await provider.getTransactionCount(sender);
+
+  while (sentProofRounds++ < proofs) {
```
folex, line 61: Please let's avoid writing such code. In-place self-increments are super hard to understand.

integration-tests/src/commitment.test.ts
```diff
@@ -19,6 +34,57 @@ const DEFAULT_TEST_TIMEOUT = 180000;
...
+          },
+        ),
+      );
+    }
+
+    await Promise.all(txs.map((tx) => tx.wait(DEFAULT_CONFIRMATIONS)));
+
+    let newEpoch;
+    do {
+      await new Promise((resolve) => setTimeout(resolve, NEXT_EPOCH_INTERVAL));
```
folex, line 80: you can instead retrieve new blocks, and get timestamp from them

integration-tests/src/commitment.test.ts
```diff
@@ -54,27 +127,99 @@ describe(
...
-        capacityContract,
+      const removeCommitmentEvent = checkEvent(
         capacityContract.filters.CommitmentRemoved,
-        1,
-        removeCommitmentTx,
+        removeCommitmentReceipt,
       );
 
-      expect(events.map((e) => e.args)).toEqual([[commitmentId]]);
+      console.log(removeCommitmentEvent.args.commitmentId);
```
folex:
```suggestion
      console.log("commitment removed: ", removeCommitmentEvent.args.commitmentId);
```

integration-tests/src/helpers.ts
folex, line 1: Could you please add return type to `registerMarketOffer`, so it is possible to understand what it returns?

eg rn it returns `appCID`, and there are no AppCIDs in Offer... So I'm confused. With explicit return type, it would be much easier to understand the intention.

integration-tests/src/helpers.ts
```diff
@@ -42,6 +43,7 @@ export async function registerMarketOffer(
   assert(lastMarketOffer, "There is no market offer");
 
   return {
+    appCID,
```
folex, line 46: Why do you need appCID here? Did you mean `metadata`?

integration-tests/src/commitment.test.ts
```diff
@@ -54,27 +127,99 @@ describe(
...
+
+      const [currentEpoch1] = await Promise.all([coreContract.currentEpoch()]);
+      console.log(currentEpoch1, "currentEpoch1");
+
+      const currentState = await capacityContract.getCommitment(commitmentId);
+      const [currentEpoch2] = await Promise.all([coreContract.currentEpoch()]);
+      console.log(currentEpoch2, "currentEpoch2");
+      expect(currentState.status).toEqual(BigInt(CCStatus.Active));
+
+      console.log("Waiting for the fail...");
```
folex:
```suggestion
      console.log("Waiting for CC to fail without proofs...");
```

integration-tests/src/commitment.test.ts
```diff
@@ -54,27 +127,99 @@ describe(
...
+      await skipEpoch(provider, epochDuration, 2);
```
folex: I think this test should rely on `maxFailedRatio`.

Otherwise, it might be that one bug hides another – and test passes.

This test should discern two situations:
1. where `Active CU count` is higher than `maxFailedRatio`, and will fail immediately no matter if sent proofs or not
2. where maxFailedRatio was reached because CC didn't submit any proofs for several epochs

integration-tests/src/commitment.test.ts
```diff
@@ -54,27 +127,99 @@ describe(
...
+      const collateralPerUnit = await capacityContract.fltCollateralPerUnit();
+
+      console.log("Depositing collateral...");
+      const depositCollateralReceipt = await capacityContract
+        .depositCollateral([commitmentId], {
+          value: collateralPerUnit,
+        })
+        .then((tx) => tx.wait(DEFAULT_CONFIRMATIONS));
+
+      // TODO: move to constant
```
folex: remove this TODO? seems like you did move it to const

integration-tests/src/commitment.test.ts
```diff
@@ -141,6 +280,163 @@ describe(
 
       const nextStatus = await capacityContract.getCommitment(commitmentId);
       expect(nextStatus.status).toEqual(BigInt(CCStatus.Failed));
```
folex: So you wait for `CC_DURATION_DEFAULT` and CC becomes `Failed` – that doesn't make sense to me.

CC becomes failed not after `CC_DURATION_DEFAULT`, but after `MAX_FAILED_RATIO / activeCCCount` epochs, right?

integration-tests/src/commitment.test.ts
```diff
@@ -141,6 +280,163 @@ describe(
...
+      const finishCommitmentTx = await capacityContract
+        .finishCommitment(commitmentId)
+        .then((tx) => tx.wait(DEFAULT_CONFIRMATIONS));
+
+      const finishCommitmentEvent = checkEvent(
+        capacityContract.filters.CommitmentFinished,
+        finishCommitmentTx,
+      );
+      assert(finishCommitmentEvent, "Not Finish commitment event");
+      expect(finishCommitmentEvent.args).toEqual([commitmentId]);
```
folex: could you also please withdraw collateral and rewards here?

this way, the scenario would be complete

integration-tests/src/commitment.test.ts
```diff
@@ -141,6 +280,163 @@ describe(
...
+      expect(currentState.status).toEqual(BigInt(CCStatus.Active));
+
+      const cuId = registeredOffer.peers[0]?.unitIds[0];
+      assert(cuId, "cuID not defined");
+
+      console.log("Sending proofs...");
+
+      await sendProof(signerAddress, cuId, 2, 3);
+
+      // const submitProofTx = await capacityContract.submitProof(
```
folex: doesn't work? if you need any help with this, let me know

integration-tests/src/deal.test.ts
```diff
@@ -173,6 +191,324 @@ describe(
...
+        registeredOffer.peers.map((p) => p.peerId),
+      );
+
+      console.info("Deposit collateral for all sent CC...");
+      await depositCollateral(capacityContract, commitmentIds);
+      await skipEpoch(provider, epochDuration, 1);
+
+      for (const ccId of commitmentIds) {
+        const status: CCStatus = Number(await capacityContract.getStatus(ccId));
+        assert(status === CCStatus.Active, "Status is not active");
```
folex:
```suggestion
        assert(status === CCStatus.Active, "CC is not active");
```

integration-tests/src/deal.test.ts
```diff
@@ -173,6 +191,324 @@ describe(
...
+      }
+
+      console.log("---- Deal Creation ----");
+      const minWorkersDeal = 1;
+      const targetWorkersDeal = 1;
```
folex: gotta check that on different number of targetWorkers

integration-tests/src/deal.test.ts
```diff
@@ -173,6 +191,324 @@ describe(
...
+      const matchDealTx = await marketContract.matchDeal(
+        dealId,
+        [registeredOffer.offerId],
+        [cuIds],
+      );
+
+      await matchDealTx.wait(DEFAULT_CONFIRMATIONS);
+      //   TODO: check further.
+
+      const [CUMatchedEvent] = await checkEvents(
```
folex: you catch only a single event here. But what if you change `targetWorkers` to any other number, like 3? 

Gotta catch them all! And make sure you catch exactly the same number of events as `targetWorkers`

integration-tests/src/deal.test.ts
```diff
@@ -173,6 +191,324 @@ describe(
...
+        CUMatchedEvent.args.appCID,
+      ]).toEqual([
+        peer.peerId,
+        dealId,
+        registeredOffer.peers[0]?.unitIds[0],
+        [dealAppCID.prefixes, dealAppCID.hash],
+      ]);
+
+      console.log("Remove compute unit participating in deal...");
+      const cuID = registeredOffer.peers[0]?.unitIds[0];
```
folex: you work on a single peer, but your test is parametrised by `targetWorkers`. Gotta work with arrays of peers and events.

integration-tests/src/deal.test.ts
```diff
@@ -173,6 +191,324 @@ describe(
...
+      );
+      assert(removeComputeUnitEvent, "No remove compute unit event");
+      expect(removeComputeUnitEvent.args).toEqual([
+        CUMatchedEvent.args.unitId,
+        dealId,
+        CUMatchedEvent.args.peerId,
+      ]);
+      const dealStatus: DealStatus = Number(await dealContract.getStatus());
+      console.log(dealStatus, "deal status");
+      expect(dealStatus === DealStatus.INSUFFICIENT_FUNDS);
```
folex: Why? That doesn't make sense. It should become `NOT_ENOUGH_WORKERS`.

Provider removed 1 CU/peer from the Deal, it shouldn't cause INSUFFICIENT_FUNDS, right?

integration-tests/src/helpers.ts
```diff
@@ -15,6 +15,7 @@ export async function registerMarketOffer(
   );
 
   console.log("Register Provider by setProviderInfo...");
+  const appCID = randomCID();
```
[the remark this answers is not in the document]
akim-bow: True, thanks

integration-tests/src/helpers.ts
akim-bow: If i add return type there it won't become cleaner. Intention here is to return various artifacts which are left after some transactions. Probably the whole function requires complete refactoring.

integration-tests/src/helpers.ts
akim-bow: I don't need that, yeah.

integration-tests/src/helpers.ts
folex: with return type, you won't be able to "return various artifacts", and 'll be forced to refactor ;)

akim-bow: I'll just leave TODO here to remember that. Return type won't help, it would be just dublicate for return object without practical use.

folex: wanna bet? :) Return types make you think stricter. It's a disgrace that TS allows to omit them.

integration-tests/src/commitment.test.ts
akim-bow: I've added additional scenario for withdrawal

akim-bow: `Long-term CC with reward withdrawals`

integration-tests/src/commitment.test.ts
```diff
@@ -54,27 +127,99 @@ describe(
...
+      console.log("Waiting for the fail...");
+      await skipEpoch(provider, epochDuration, 2);
+      const currentStateAfterSomeEpoches =
+        await capacityContract.getCommitment(commitmentId);
+      expect(currentStateAfterSomeEpoches.status).toEqual(
+        BigInt(CCStatus.Failed),
+      );
+    });
 
+    test("CC ends after duration", async () => {
```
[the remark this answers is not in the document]
akim-bow: Right, fixed

integration-tests/src/deal.test.ts
akim-bow: I will add TODO for that.

integration-tests/src/deal.test.ts
akim-bow: Pointed in TODO

integration-tests/src/deal.test.ts
```diff
@@ -173,6 +191,324 @@ describe(
...
+        CUMatchedEvent.args.unitId,
+        CUMatchedEvent.args.appCID,
+      ]).toEqual([
+        peer.peerId,
+        dealId,
+        registeredOffer.peers[0]?.unitIds[0],
+        [dealAppCID.prefixes, dealAppCID.hash],
+      ]);
+
+      console.log("Remove compute unit participating in deal...");
```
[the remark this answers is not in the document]
akim-bow: Added setWorker call

integration-tests/src/helpers.ts
akim-bow: In general i agree that return type should be defined. In this particular case there is no use of return type. Return type simply copies unstructured object and doesn't provide any befefit to code strictestness. But yeah, for you as a reviewer would be easier to look at the code where everything is explicitly typed.

integration-tests/src/commitment.test.ts
```diff
@@ -52,394 +28,574 @@ async function sendProof(
...
+    await sendProof(signerAddress, cuId, MAX_ALLOWED_PROOFS, 1);
+
+    const currentStateAfterSentProofs =
+      await capacityContract.getCommitment(commitmentId);
+    expect(currentStateAfterSentProofs.status).toEqual(BigInt(CCStatus.Active));
+
+    console.log("Sending excessive amount of proofs...");
+    await expect(
+      sendProof(signerAddress, cuId, MAX_ALLOWED_PROOFS * 2, 2),
+    ).rejects.toBeDefined();
```
folex, line 442: please check specifics of the error. I hope that's possible. If it is an unexpected revert like overflow or something, tests must catch it and report.

PR #348, opened 2024-03-14
fix: Vesting.add

ValeryAntopol:


elshan-eth:
I merged it in audit-fix branch

PR #349, opened 2024-03-15
fix!: deal rewards

elshan-eth:
- mv capacity const to core (affect CLI and NOX)
- fix by audit-report
- fix vesting and reward pool calculation


src/deal/Deal.sol
```diff
@@ -374,94 +200,178 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
...
+        }
+
+        uint256 currentWorkerCount = getWorkerCount();
         uint256 pricePerWorkerEpoch_ = pricePerWorkerEpoch();
+        snapshot =
+            DealSnapshot.init(dealStorage, currentEpoch, commitEpoch, isEnded, pricePerWorkerEpoch_, currentWorkerCount);
 
-        uint256 workerCount = getWorkerCount();
+        uint256 lastCommitedEpoch = dealStorage.lastCommitedEpoch;
+        if (commitEpoch <= lastCommitedEpoch) {
```
ValeryAntopol, line 288: why is this check is not at the start of the method?

src/deal/Deal.sol
```diff
@@ -374,94 +200,178 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
...
+                snapshot.setLockedBalance(snapshot.getLockedBalance() + amount); // record locked balance for rewards
+                snapshot.setGapsEpochCount(snapshot.getGapsEpochCount() + (commitEpoch - maxPaidEpoch)); // write gaps
+            }
+        } else {
+            // commitEpoch <= maxPaidEpoch, this means that we we need to record only active Epochs and write off balances. We don't have a gaps.
+            uint256 amount = (commitEpoch - lastCommitedEpoch) * pricePerWorkerEpoch_ * currentWorkerCount;
 
-        emit DealEnded(currentEpoch);
+            snapshot.setTotalBalance(snapshot.getTotalBalance() - amount); // write off balance
+            snapshot.setLockedBalance(snapshot.getLockedBalance() + amount); // record locked balance for rewards
```
ValeryAntopol, line 312: maybe decouple this part into a function? this code is doubled in the other branch of if

elshan-eth: I think we’ll do it later

src/deal/Deal.sol
elshan-eth: Because before we init snapshot

integration-tests/src/commitment.test.ts
```diff
@@ -235,16 +235,17 @@ describe("Capacity commitment", () => {
...
     const HACKY_WITHDRAW_EPOCHS = withdrawEpochs * 2n;
     await skipEpoch(provider, epochDuration, HACKY_WITHDRAW_EPOCHS);
 
+    // TODO: Add error class
     await expect(
       capacityContract
         .finishCommitment(commitmentId)
         .then((tx) => tx.wait(DEFAULT_CONFIRMATIONS)),
-    ).rejects.toThrow("For finish commitment all units should be exited");
+    ).rejects.toThrow("(unknown custom error)");
```
folex, line 248: 😭

akim-bow: Я уже готовлю фикс для этого в след пулике. Это чтоб тесты прошли

PR #376, opened 2024-03-20
fix: use load-or-create-provider instead of manual decide

AlcibiadesCleinias:


subgraph/src/mappings/core.ts
```diff
@@ -32,21 +37,16 @@ export function handleInitialized(event: Initialized): void {
 export function handleWhitelistAccessGranted(
   event: WhitelistAccessGranted,
 ): void {
-  let provider = Provider.load(event.params.account.toHexString());
-  // TODO: HANDLE PROPERLY
-  if (provider !== null) {
-    provider.approved = true;
-    provider.save();
-  }
+  log.info('TODO DEBUG {}', [formatAddress(event.params.account)])
```
justprosh, line 40: remove debug logs

subgraph/src/mappings/market.ts
```diff
@@ -125,12 +120,15 @@ export function handleMarketOfferRegistered(
 
 // It updates Peer and Offer.
 export function handleComputeUnitCreated(event: ComputeUnitCreated): void {
+  log.info('TODO DEBUG handleComputeUnitCreated', [])
```
justprosh, line 123: ditto

subgraph/src/models.ts
```diff
@@ -29,19 +29,23 @@ export const REMOVED_EFFECTOR_INFO_DESCRIPTION = "Removed";
...
 ): Provider {
-  let entity = Provider.load(providerAddress);
+  const providerAddressSerialized = providerAddress;
+  let entity = Provider.load(providerAddressSerialized);
 
   if (entity == null) {
-    entity = new Provider(providerAddress);
-    entity.registered = false;
+    entity = new Provider(providerAddressSerialized);
+    entity.registeredAt = ZERO_BIG_INT;
```
justprosh, line 45: why is it zero here?

justprosh: ok, you set it outside of create but i am not sure why bool is not enough

subgraph/src/mappings/core.ts
AlcibiadesCleinias: fffff

subgraph/src/models.ts
AlcibiadesCleinias: I believe that info about when registered is more useful than just registered

justprosh: kk

subgraph/src/models.ts
```diff
@@ -29,19 +29,25 @@ export const REMOVED_EFFECTOR_INFO_DESCRIPTION = "Removed";
...
 // @notice Those unregistered providers does not count in total providers.
-export function createOrLoadUnregisteredProvider(
+export function createOrLoadProvider(
   providerAddress: string,
+  timestamp: BigInt,
 ): Provider {
-  let entity = Provider.load(providerAddress);
+  // 'I do not repeat, I do not repeat'. One more lowercase in case if someone will
+  //  forget to use it in the future.
+  const providerAddressSerialized = providerAddress.toLowerCase();
```
justprosh, line 42: just do formatAddress here and accept provider address in any form

AlcibiadesCleinias: yeah, that would be better

AlcibiadesCleinias: oh, no, I forgot that formatAddress is for indexer "Address" type

AlcibiadesCleinias: so, we should not use formatAddress

justprosh (approved):

PR #384, opened 2024-03-25
feat(ts-client): remove explorer client from ts-client

akim-bow:


ts-client/src/dealCliClient/dealCliClient.ts
```diff
@@ -48,23 +48,42 @@ export class DealCliClient {
...
-      return serializeOfferDetail(data.offer, this._serializationSettings)
+      return serializeOfferDetail(data.offer, this._serializationSettings);
+    }
+    return null;
+  }
+
+  /*
+   * @notice [Figma] Offer.
+   */
+  async getOfferLikeExplorer(offerId: string): Promise<OfferDetail | null> {
```
AlcibiadesCleinias, line 59: what is it? have you updated with main: it seemed to me that it is already implemented and delivered: https://github.com/fluencelabs/deal/blob/main/ts-client/src/dealCliClient/dealCliClient.ts#L45

AlcibiadesCleinias: bad naming, though.

ts-client/src/utils/serializers.ts
```diff
@@ -21,15 +21,7 @@ export function tokenValueToRounded(
   const formatted = ethers.formatUnits(value, decimals);
   const parsed = parseFloat(formatted).toFixed(toFixed);
   if (parsed == "0." + "0".repeat(toFixed)) {
-    return "0"
+    return "0";
   }
-  return parsed
-}
-
-// Convert human-readable value to token value with its decimals (e.g. WEI in case of ETH-like tokens).
```
AlcibiadesCleinias, line 29: why deleted?

PR #397, opened 2024-03-29
feat!: oracle role CHAIN-444

crystalbit:
## ChangeLog
- Changes around setFLTPrice - it was onlyOwner, now it can be triggered only with oracle
- added oracle role (`setOracle(address) onlyOwner`, `oracle() view`)


src/core/interfaces/ICapacityConst.sol
```diff
@@ -15,6 +15,10 @@ interface ICapacityConst is IEpochController {
     /// @param newValue The new value of the constant
     event CapacityConstantUpdated(CapacityConstantType constantType, uint256 newValue);
 
+    /// @dev Emitted then new oracle address is set
```
AlcibiadesCleinias, line 18: when*

src/core/CapacityConst.sol
```diff
@@ -217,8 +222,9 @@ contract CapacityConst is ICapacityConst, OwnableUpgradableDiamond, EpochControl
     // #endregion ------------------ External View Functions ------------------
 
     // #region ------------------ External Mutable Functions ------------------
-    function setFLTPrice(uint256 fltPrice_) public onlyOwner {
+    function setFLTPrice(uint256 fltPrice_) public {
```
AlcibiadesCleinias, line 225: not sure if we want. But from code prospective it is better to be flt agnostic in contract namings. Rly not sure

AlcibiadesCleinias: ok, I reread seems we want to leave as it is

src/core/interfaces/ICapacityConst.sol
crystalbit: oh :) ty

src/core/CapacityConst.sol
crystalbit: I think so too, FLT is already everywhere and isn't going to be changed

AlcibiadesCleinias (approved):

PR #401, opened 2024-04-01
fix: Light refactoring

elshan-eth:


Makefile
```diff
@@ -10,6 +10,14 @@ LOCAL_CHAIN_BLOCK_MINING_INTERVAL ?= 1
 verify-command: ## Verify command
 	@command -v $(program) > /dev/null || (echo "\033[0;31m$(program) is not installed. Please install $(program) and try again.\033[0m" && exit 1)
 
+fmt:
```
AlcibiadesCleinias, line 13: strange, in makefile we used to command in src, ts-client, etc. mb better fmt-contracts.

Makefile
```diff
@@ -38,9 +46,15 @@ build-npms: ## Build all npms: subgraph and ts-clients
 
 build-all: build-contracts build-npms ## Build contracts and npms
 
+run-contract-tests:  ## Test for solidity contracts & ts-clients
```
AlcibiadesCleinias, line 49: not true - you test only contracts here.

Also, pls `run-tests` rename to `test-contracts`
probably add separate `test-ts-clients` or smthing

PR #403, opened 2024-04-02
feat: cliClient get offers and offer ids by provider[CHAIN-448, CHAIN-449]

AlcibiadesCleinias:


ts-client/src/dealCliClient/dealCliClient.ts
```diff
@@ -65,4 +105,28 @@ export class DealCliClient {
...
+      limit: paginator.limit,
+      orderBy: order.orderBy,
+      orderType: order.orderType,
+    });
+    return (
+      data.offers.map((offer) => {
+        return {
+          id: offer.id,
+        };
+      }) || []
```
crystalbit, line 129: `|| []` does nothing. Maybe you wanted `(data.offers ?? []).map` or `data.offers?.map(...) ?? []`

AlcibiadesCleinias: yeah, probably need to delete it, thanks

crystalbit (approved):

PR #412, opened 2024-04-03
feat!: Next release

elshan-eth:
- add oracle role to set the price
- refactoring repo structure and test for performance 
- fix bugs from the audit report 
- removeСUFromCC had a bug with return units from the deal. Now CLI needs to make returnComputeUnitFromDeal before removeCUFomCC. https://github.com/fluencelabs/cli/pull/897


foundry.toml
```diff
@@ -23,3 +27,11 @@ stage = "https://ipc-stage.fluence.dev"
 dar = "https://ipc.dar.fluence.dev"
 kras = "https://ipc.kras.fluence.dev"
 mumbai = "https://rpc.ankr.com/polygon_mumbai"
+
+[profile.test]
+src = "test"
+out = "test-out"
```
AlcibiadesCleinias, line 33: `out-test` pls, coz we will have out close to out-test in dir tree

src/core/modules/capacity/interfaces/ICapacity.sol
```diff
@@ -144,6 +144,7 @@ interface ICapacity {
         uint256 snapshotEpoch;
         uint256 activeUnitCount;
         uint256 nextAdditionalActiveUnitCount;
+        mapping(uint256 => uint256) successCountByEpoch;
```
elshan-eth, line 147: This a collision!!!!

src/core/modules/capacity/interfaces/ICapacity.sol
```diff
@@ -136,6 +135,7 @@ interface ICapacity {
         uint256 filledRemainingFailedUnitsInLastEpoch;
         uint256 exitedUnitCount;
         uint256 totalSlashedCollateral;
+        mapping(bytes32 => bool) isUnitExited;
```
elshan-eth, line 138: This a collision!!!!

elshan-eth: They are shifting the next var slot in the Commitment structure.

PR #419, opened 2024-04-05
fix!: separate withdraw - breaking changes (N02)

elshan-eth:
- breaking changes for CLI
- to withdraw CC reword, CLI needs to call fixReward before withdrawReward
- to withdraw CC collateral, CL needs to call withdrawCollateral after finishCommitment


AlcibiadesCleinias:
also it adds possibility for anyone (prev. only provider) to removeCUfromCC if conditions are met

elshan-eth:
> also it adds possibility for anyone (prev. only provider) to removeCUfromCC if conditions are met
I didn't quite get it. Now any address can call removeCUfromCC. We have no restrictions for this

PR #426, opened 2024-04-16
feat: storage checker

crystalbit:
storage checker script for ci/cd


storage-checker/storage-layout.json
```diff
@@ -0,0 +1,818 @@
+{
```
AlcibiadesCleinias, line 1: `storage-layout-dump.json` ?

storage-checker/compare.js
```diff
@@ -0,0 +1,66 @@
+const fs = require("fs");
+const path = require("path");
+const { compareSlots } = require("./compare-slots");
+
+const outDir = path.join(__dirname, "..", "out");
+const storageLayoutFile = path.join("ci/", "storage-layout.json");
```
AlcibiadesCleinias, line 6: `prevStorageLayoutFile` in accordance with code below

storage-checker/storage-layout.json
AlcibiadesCleinias: prev-storage-layout.json

AlcibiadesCleinias (approved):
The only thing left is to adopt script for deleted storage slots.

We discussed that it might be useful in contract to have agreement on prefix naming for deleted slots in conjunction with script. E.g.

```
contract Foo {
uint constant slotName = 1;
}

contract UpgradedFoo {
uint constant __deprecatedStorageSlot__slotName;
}
```

and thus, in the script we could catch such names and also ignore "changes" (test failes)

PR #449, opened 2024-05-27
feat: oracle

crystalbit:
Oracle role for setFLTPrice

### DIFF
--- a/test/core/CapacityConst.t.sol
+++ b/test/core/CapacityConst.t.sol
@@ -58,7 +58,8 @@ contract CpacityConstTest is TestWithDeployment {
             maxFailedRatio: 100,
             difficulty: 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF,
             initRewardPool: 1500 ether,
-            randomXProxy: address(0x123)
+            randomXProxy: address(0x123),
+            oracle: address(456)
         });
 
         _initCapacityConst(args);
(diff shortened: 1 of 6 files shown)


test/core/CapacityConst.t.sol
```diff
@@ -201,6 +207,16 @@ contract CpacityConstTest is TestWithDeployment {
         uint256 newPrice = 4 * PRECISION;
 
         // #region set price in first epoch
+        vm.prank(NOT_AN_OWNER);
+        vm.expectRevert(abi.encodeWithSelector(OwnableUnauthorizedAccount.selector, NOT_AN_OWNER));
+        capacityConst.setOracle(address(12345678));
+
+        capacityConst.setOracle(address(12345678));
```
AlcibiadesCleinias, line 214: rm it pls

test/core/CapacityConst.t.sol
```diff
@@ -201,6 +207,16 @@ contract CpacityConstTest is TestWithDeployment {
...
+        vm.expectRevert("Only oracle can set FLT price");
+        capacityConst.setFLTPrice(newPrice);
+
+        capacityConst.setOracle(args.oracle);
```
AlcibiadesCleinias, line 218: probably this line should not be here

AlcibiadesCleinias (changes requested):
Everything seems good, but as I understand in tests some line is better to remove

AlcibiadesCleinias:
I added clean cache to restart the e2e tests

AlcibiadesCleinias (approved):

PR #450, opened 2024-05-27
feat: CC-filters-CHAIN-476

AlcibiadesCleinias:
# ChangeLog
* refactor package command to test matcher also (prev it tests only appended custom test from @shamsartem )
   * refactor package layout a bit
   * readme bump

# Breaking Change
* DealExplorerClient: added new status filter to `getCapacityCommitments`
  * added explorer tests to CI separately from matcher (another command, the same workflow)
    * added test for dealExplorerClient `getCapacityCommitments` filter by status
* the same filter for Provider.CC page [CHAIN-547]
* the same filter for Peer.CC page [CHAIN-549]

    
# ToDo
- [ ] redeploy subgraph


ts-client/src/dealExplorerClient/types/filters.ts
```diff
@@ -36,10 +46,17 @@ export interface ChildEntitiesByProviderFilter {
   status?: ProviderChildEntityStatusFilter | undefined;
 }
 
+// @param status: do not specify status to retrieve all
+export interface CapacityCommitmentsByProviderFilter {
```
folex, line 50: Can you also add `ByStaker` / `ByDelegator`? Not sure if it's small enough to do in the same PR

InversionSpaces: Did `ByDelegator`

.github/workflows/tests-clients-integration.yml
```diff
@@ -115,7 +115,10 @@ jobs:
         env:
           CHAIN_RPC_IMAGE: "chain-rpc"
           CHAIN_DEPLOY_SCRIPT_IMAGE: "chain-deploy-script"
-          EPOCH_DURATION: "15" # seconds, tests are waiting for epoch change in real time
+          # Following parameters are expected in integration tests
+          MAX_FAILED_RATIO: "4"
+          MIN_DURATION: "1"
+          EPOCH_DURATION: "20" # seconds, tests are waiting for epoch change in real time
```
folex, line 121: If you set it to 1 seconds, will tests pass quicker?

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+          [providerFixture],
+          marketContract,
+          WAIT_CONFIRMATIONS,
+        );
+
+        console.log("--- Check that there are no CCs at the start. ---");
+
+        {
+          const fetchedCCs = await ccFetcher.fetch();
+          expect(fetchedCCs.length).toEqual(0);
```
folex, line 126: add msg maybe? ie `there shouldn't be any CCs past ${ccFetcher.createdAt}, found ${fetchedCCs}`

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        }
+
+        const createdCCIds = await createCommitmentForProviderFixtures(
+          [providerFixture],
+          capacityContract,
+          WAIT_CONFIRMATIONS,
+          duration,
+        );
+
+        await delay(SUBGRAPH_MAX_INDEXING_TIME, "Wait for subraph to index");
```
folex, line 136: Hmm, maybe wait in a loop with short pauses and check this on every iteration? Basically, polling.
```
          const fetchedCCs = await ccFetcher.fetch(["waitDelegation"]);
          expect(fetchedCCs.length).toEqual(createdCCIds.length);
```

Constant pauses have their downsides:
- If CI worker slows down, tests fail randomly
- Tests never pass faster even if the system performance increases
- Pauses are always pessimistic, so they tend to increase over time. And being used in several places, that causes major slowdowns which are super hard to debug.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        await epochSkipper.skip();
+
+        console.log("Deposit collateral for created CCs: ", createdCCIds);
+        await depositCollateral(
+          createdCCIds,
+          capacityContract,
+          WAIT_CONFIRMATIONS,
+        );
+
+        await delay(SUBGRAPH_MAX_INDEXING_TIME, "Wait for subraph to index");
```
folex, line 175: ditto: wdyt about polling instead of constant delay?

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+          expect(epochDuration * 1000n).toBeGreaterThanOrEqual(
+            SUBGRAPH_MAX_INDEXING_TIME * 2,
+          );
+
+          const maxFailedRatio = await coreContract.maxFailedRatio();
+          /**
+           * MAX_FAILED_RATIO should be greater than capacity duration bc
+           * this test wont send CC proofs and don't expect CC to fail.
+           */
+          expect(maxFailedRatio).toBeGreaterThan(CAPACITY_DEFAULT_DURATION);
```
folex, line 225: Hmm, should it be `CAPACITY_DEFAULT_DURATION * coreContract.minRequiredProofs()`?

- `maxFailedRatio` unit is number of proofs
- `CAPACITY_DEFAULT_DURATION` unit is number of epochs

So to compare apples to apples, ie number of proofs to number of proofs, gotta multiply `CAPACITY_DEFAULT_DURATION` by `minRequiredProofs`, I would think.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+      });
+
+      /**
+       * Function to create a capacity commitment and go through the process of activating it.
+       * This is common start for some tests.
+       *
+       * @param duration - The duration of the commitment.
+       * @return An array containing the IDs of created capacity commitments, provider fixture, and deal fixture.
+       */
+      async function createCCToActive(
```
folex, line 89: Maybe move this function to a separate file? It's pretty big, and IMO it's nice to have only test suite code in a file, and all utils separately, so no one every confuses util function impl with test scenario. If you're OK with that.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        );
+
+        console.log("--- Check that there are no CCs at the start. ---");
+
+        {
+          const fetchedCCs = await ccFetcher.fetch();
+          expect(fetchedCCs.length).toEqual(0);
+        }
+
+        const createdCCIds = await createCommitmentForProviderFixtures(
```
folex, line 129: Why is it plural? Shouldn't it create a single CC?

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+      /**
+       * Function to create a capacity commitment and go through the process of activating it.
+       * This is common start for some tests.
+       *
+       * @param duration - The duration of the commitment.
+       * @return An array containing the IDs of created capacity commitments, provider fixture, and deal fixture.
+       */
+      async function createCCToActive(
+        duration: number,
+      ): Promise<[string[], ProviderFixtureModel, DealFixtureModel]> {
```
folex, line 91: Function is called `create capacity commitment`, ie a single one. But you it returns an array of CC IDs – why?

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
folex: I think it makes sense to assert that length equals to 1, and take that single CC, and work with a singular value

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        await sign(capacityContract.finishCommitment, chosenCC);
+
+        await delay(SUBGRAPH_MAX_INDEXING_TIME, "Wait for subraph to index");
+
+        // Target check on chain.
+        {
+          const statusCCFromChain = await capacityContract.getStatus(chosenCC);
+          expect(Number(statusCCFromChain)).toEqual(CommitmentStatus.Removed);
+        }
+        // Target check in client, we removed only 1 CC out of 2.
```
folex, line 290: I don't understand where you create 2 CCs, I only see 1 CC being created on line 228

.github/workflows/tests-clients-integration.yml
InversionSpaces: Lowering this number will decrease tests run time, but epoch duration should be enough for a transaction + the graph indexing, so 1 second wont do probably. Also the smaller the number, the more flaky tests will be

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+import { delay, EpochSkipper } from "../utils/misc.js";
+import { sign } from "../utils.js";
+import { CCFetcher } from "../utils/misc.js";
+
+// Empirically measured time for subgraph indexing on 4CPU, 8Gb mem.
+const DEFAULT_SUBGRAPH_TIME_INDEXING = 300000;
+const TESTS_TIMEOUT = 120000 + 30000 + DEFAULT_SUBGRAPH_TIME_INDEXING;
+const CAPACITY_DEFAULT_DURATION = 3; // if 1 epoch 20 sec -> 60 sec.
+const CAPACITY_FAIL_DURATION = 5; // if 1 epoch 20 sec -> 100 sec.
+const SUBGRAPH_MAX_INDEXING_TIME = 10000; // 10 [sec].
```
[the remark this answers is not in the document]
InversionSpaces: Frankly speaking, I don't know what is `DEFAULT_SUBGRAPH_TIME_INDEXING` and why it is added to `TESTS_TIMEOUT`

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        {
+          const fetchedCCs = await ccFetcher.fetch(["waitDelegation"]);
+          expect(fetchedCCs.length).toEqual(createdCCIds.length);
+        }
+
+        console.log(
+          "--- Check that status is WaitStart after depositing collateral. ---",
+        );
+        /**
+         * NOTE: We skip to the start of an epoch so that
```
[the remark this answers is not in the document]
InversionSpaces: Well, yes, we skip to the start of the next epoch, but it does not matter a lot, test just have to start at the beginning of any epoch.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: But I would not say it is a utility function, it is basically part of the test code, it has a lot of `expect`. It is just common between two tests, so I moved it to a separate function.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: It creates two CCs. Tests are written so that they don't care how many CCs are created (if at least one is created). In the first however at the end only one CC is removed and others are not, so it is an additional test check that other CCs do not become removed.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: Because it creates more than one CC. I will it plural in the name

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: There were created two CCs

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: AFAIK, no `maxFailedRatio` is number of fails per CU, where "fail" is "CU did not send enough proofs in one epoch" (no matter how many did it send). e.g. see comment in `LibCapacity.sol`:
```
        // maxFailCount is a number of Epochs when units not send proofs
        // if on unit not send proof per one epoch, it's mean that it's one fail
        // maxFailedRatio is the ratio of failed attempts for all units
        uint256 maxFailCount = maxFailedRatio_ * peer.unitCount;
```
or see function `calculateNextFailedCCEpoch`, it does not take into account `minRequiredProofs` at all.

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: With polling:
- If worker slows down, tests still can fail randomly because epoch could pass while the graph is indexing and tests do not expect that
- Tests still wont pass much faster on more performant worker because they are still waiting for epoches to pass

Given the fact that tests anyway have to wait for epoches to pass and for indexer to keep up with epoches, I don't see much benefits in polling. 

Still rewrote to polling.

DieMyst:
utACK

DieMyst (approved):

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
```diff
@@ -262,93 +298,95 @@ describe(
...
-        const [createdCCIds] = await createCCToActive(CAPACITY_FAIL_DURATION);
+        const [createdCCIds] = await createCCsToActive(
+          CAPACITY_FAIL_DURATION_EPOCHES,
+        );
 
         console.log("--- Check that status is Failed after CC is failed. ---");
 
         // Wait for CC to fail, thus, wait for CAPACITY_FAIL_DURATION epochSkipper to pass.
-        await epochSkipper.skip(CAPACITY_FAIL_DURATION);
+        await epochSkipper.skip(CAPACITY_FAIL_DURATION_EPOCHES);
```
folex, line 370: EPOCHS*

folex (approved):

integration-tests-clients/src/networkExplorer/dealExplorerClient.test.ts
InversionSpaces: Fixed in the next [PR](https://github.com/fluencelabs/compute-marketplace/pull/484)

PR #461, opened 2024-07-04
chore: Add prettier action

InversionSpaces:


.github/workflows/prettier.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
+    
+      - name: Run Prettier
+        run: npx prettier@3.3.2 --write .
+    
+      - name: Commit Prettier Changes
```
folex, line 33: I'm not sure it's a good idea to commit. Just check is enough, let dev be responsible for this kind of thing

folex (approved):
All good except for the automatic commit of prettier.

That would generate a lot of noise, better let humans handle the mutation, and robots – verification.

DieMyst (approved):

PR #476, opened 2024-07-11
feat(CapacityCommitments): optimize collateral withdraw

ValeryAntopol:
Includes:
* reworked fail calculation
* reworked slashing calculation
* stopping deal rewards for CUs associated with failed commitments


ValeryAntopol:
closed in favor of #508

PR #483, opened 2024-07-18
fix: exit

crystalbit:
Mostly cherry-picked from #412 

# ChangeLog
- Fix bug with exit, fix spelling & fix failedEpoch formula, fix success count in failsInFirstEpoch calculation
- Matching logic fix
- save maxProofsPerEpoch fix
- fix submit proof with inactive units
- fix small balance problem
- fix bug with status


src/core/Matcher.sol
```diff
@@ -129,12 +129,15 @@ abstract contract Matcher is Offer, IMatcher {
...
                 if (
+                    computeUnit.deal != address(0)
                     // Blacklisted provider we filtered out above via: !deal.isProviderAllowed(offer.provider).
-                    providersAccessType != IConfig.AccessType.WHITELIST
-                        && (
-                            computeUnit.deal != address(0) || peer.commitmentId == bytes32(0x000000000)
-                                || LibCapacity.getStatus(peer.commitmentId) != ICapacity.CCStatus.Active
-                        )
+                    || (
+                        providersAccessType != IConfig.AccessType.WHITELIST
```
folex, line 135:
```suggestion
// If Deal does not specify whitelist of Providers, then do not match such peers that do not have an active Capacity Commitment
                        providersAccessType != IConfig.AccessType.WHITELIST
```

src/core/interfaces/ICapacity.sol
```diff
@@ -239,6 +246,8 @@ interface ICapacity {
     /// @param resultHash Target hash of this proof
     function submitProof(bytes32 unitId, bytes32 localUnitNonce, bytes32 resultHash) external;
 
+    function getRewardInfo(uint256 epoch) external view returns (RewardInfo memory);
```
folex, line 249:
```suggestion
    // Returns number of submitted proofs along with min and max required number of proofs
    function getRewardInfo(uint256 epoch) external view returns (RewardInfo memory);
```

src/facets/CapacityFacet.sol
```diff
@@ -273,9 +289,12 @@ contract CapacityFacet is ICapacity, DiamondReentrancyGuard {
             );
 
             uint256 unitProofCount = unitInfo.proofCountByEpoch[currentEpoch] + 1;
-            if (unitProofCount > LibCapacityConst.maxProofsPerEpoch()) {
+
+            RewardInfo storage rewardInfo = LibCapacity._getOrSaveRewardPool(currentEpoch);
+            if (unitProofCount > rewardInfo.maxProofsPerEpoch) {
                 revert TooManyProofs();
             }
+            uint256 minProofsPerEpoch_ = rewardInfo.minProofsPerEpoch;
```
folex, line 297: Probably we do not revert on TooManyProofs for `submitProofs` (ie batching proof verification) anymore

src/deal/Deal.sol
```diff
@@ -207,12 +207,15 @@ contract Deal is MulticallUpgradeable, WorkerManager, IDeal {
...
+        Status status = getStatus();
+        require(status != Status.ENDED && status != Status.INSUFFICIENT_FUNDS, "Deal is not active");
+
         if (prevWorkerCount == newWorkerCounts) {
             return;
         }
 
-        _postCommitPeriod(snapshot, newWorkerCounts);
-
         ComputeUnitPaymentInfo storage computeUnitPaymentInfo = dealStorage.cUnitPaymentInfo[computeUnitId];
```
folex, line 219:
```suggestion
        // Only write these values to ComputeUnitPaymentInfo IF we went from state NO WORKER ID to state HAS WORKER ID
        // This is because snapshotEpoch and gapsDelta are used to calculate withdraws
        // Basically, this state transition means "CU" was activated. It must happen only once per CU in a Deal, it seems.
        ComputeUnitPaymentInfo storage computeUnitPaymentInfo = dealStorage.cUnitPaymentInfo[computeUnitId];
```

ValeryAntopol:
will be moved to optimize collateral PR #476

PR #487, opened 2024-07-24
chore: Add ESLint

InversionSpaces:
Add eslint and typescript typechecking to tests, ts-client and subgraph. 
For subgraph package, typescript-eslint rules were modified to roughly match AssemblyScript.


InversionSpaces:
Note: I was not able to force vitest to run typescript typechecking on source code, so tsc is used in `npm run lint` script

.github/workflows/eslint.yaml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - name: Setup foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+
+      - name: Build contracts
```
folex, line 32: Hmm, I believe you also need to do `forge install`.

There's a target in Makefile called `install-deps`, I suggest you to use it. But rn as I see, there's typo there: it does `forge build` instead of `forge install`:
```make
install-deps:
	@make install-npms
	@make verify-command program=forge
	@forge build
	@echo "\033[0;32mSuccess! Installed forge deps.\033[0m"
```

Could you please fix it, ie change `forge build` to `forge install`? Thanks!

.github/workflows/eslint.yaml
```diff
@@ -0,0 +1,54 @@
...
+        run: make build-contracts
+
+      # Build ts-client separately so it is installed
+      # via `npm i` in integration test packages
+
+      - name: Install deps in ts-client package
+        run: npm install
+        working-directory: ts-client
```
folex, line 40: `install-npms` already does `npm install` in `ts-client`

integration-tests-clients/package.json
```diff
@@ -7,21 +7,25 @@
...
-    "lint": "prettier src --check && eslint --cache \"./src/**/*.ts\"",
+    "test": "vitest run --no-file-parallelism",
+    "lint": "tsc --noemit && eslint . --cache",
+    "fmt": "prettier 'src/**/*.ts' --write",
+    "fmt-check": "prettier 'src/**/*.ts' --check",
     "set-local-env": "EPOCH_DURATION=86400 docker compose -f ../docker/docker-compose.yml up -d --build chain-deploy-script"
   },
-  "author": "",
+  "author": "Cloudless Labs",
   "license": "ISC",
```
folex, line 20: Could you please change license to AGPL v3.0 🙏 I've missed it, it seems.

.github/workflows/eslint.yaml
InversionSpaces: Yes, but the reason to do so is mentioned in the comment above. To install current ts-client in test packages, we need to build it first.

folex: I think it would be best to have this scenario described in Makefile, so anyone can reproduce it locally.

InversionSpaces: This way `make install-npms` will basically trigger installation of contracts deps and build of contracts, you sure it is good idea?

ts-client/src/dealExplorerClient/serializers/orderby.ts
```diff
@@ -31,24 +31,22 @@ export function serializeDealShortOrderByToIndexer(
...
 ): CapacityCommitment_OrderBy {
-  if (v == "createdAt") {
-    return "startEpoch";
+  switch (v) {
+    case "createdAt":
+      return "startEpoch";
+    case "expirationAt":
+      return "endEpoch";
+    case "computeUnitsCount":
+      return "computeUnitsCount";
```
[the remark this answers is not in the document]
InversionSpaces: Why? All cases of input value are covered. Switch is exhaustive. Addition of `default` case would complicate possible updates - ts would not report non exhaustive switch.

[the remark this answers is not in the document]
crystalbit: for example case when we add more values to the type and forget to add here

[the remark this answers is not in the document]
InversionSpaces: This is exactly the use case for linter (compiler check to be precise). It will complain **on lint** (rather than in runtime) that you forgot to add a case:
image

folex (approved):

DieMyst:
utACK

DieMyst (approved):

ts-client/src/dealExplorerClient/serializers/orderby.ts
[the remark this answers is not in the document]
crystalbit: ok :)

crystalbit (approved):

PR #516, opened 2024-07-31
feat(staker): Staker API for staker page

DieMyst:
fixes CHAIN-648
fixes CHAIN-649


ts-client/src/dealExplorerClient/serializers/stakerEncoders.ts
```diff
@@ -90,3 +150,18 @@ export function serializeOrderBy(
...
+): CapacityCommitment_OrderBy {
+  switch (orderBy) {
+    case "duration":
+      return "duration";
+    case "reqCollateral":
+      return "requiredCollateral";
+    case "reward":
+      return "rewardDelegatorRate";
+    case "expectedAPR":
+      return "createdAt";
```
crystalbit, line 165: looks like a mistake

ts-client/src/dealExplorerClient/stakerClient.ts
```diff
@@ -30,6 +36,8 @@ export class StakerClient {
 
     const constants = serializeConstants(await indexer.getConstants());
 
+    constants;
```
crystalbit, line 39: if it is only for usage, better to use ts-ignore imho

ts-client/src/dealExplorerClient/utils.ts
```diff
@@ -44,6 +44,16 @@ export function calculateEpoch(
   );
 }
 
+// Duration from timestamp to number of epochs
+export function calculateDuration(
+  timestamp: number,
+  epochControllerStorageEpochDuration: number,
+): number {
+  return parseInt(
```
crystalbit, line 52: Maybe better `Math.floor`?

crystalbit: I mean instead of parseInt(.toString())

PR #524, opened 2024-08-02
chore: add vesting events

crystalbit:


src/core/Vesting.sol
```diff
@@ -88,19 +91,26 @@ library Vesting {
         }
 
         for (uint256 i = 0; i < vestingPeriodCount; i++) {
+            uint256 amount = rewardPerPeriod * (i + 1);
```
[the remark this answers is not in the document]
ValeryAntopol: That is because we are counting cumulative sums. `cumulateiveAmountIncrease` may be better name

folex, line 94: Let's call it `cumulativeAmount`, then.
```suggestion
            uint256 cumulativeAmount = rewardPerPeriod * (i + 1);
```

src/core/Vesting.sol
```diff
@@ -30,6 +30,9 @@ library Vesting {
         uint256 cumulativeAmount;
     }
 
+    event VestingScheduleItem(uint256 indexed id, uint256 indexed epoch, uint256 amount, uint256 cumulativeAmount);
+    event VestingUpdateItem(uint256 indexed id, uint256 amount, uint256 cumulativeAmount);
+
```
ValeryAntopol, line 35: I'm not sure if the indexer can use the `amount` field meaningfully. But anyway, this should be fine

src/core/Vesting.sol
```diff
@@ -88,19 +91,26 @@ library Vesting {
         }
 
         for (uint256 i = 0; i < vestingPeriodCount; i++) {
+            uint256 amount = rewardPerPeriod * (i + 1);
             // add reward to cumulativeAmount
```
folex, line 95: And I guess this comment should be rewritten, right

ValeryAntopol (approved):
utACK

folex (approved):

src/core/Vesting.sol
```diff
@@ -88,19 +91,27 @@ library Vesting {
         }
 
         for (uint256 i = 0; i < vestingPeriodCount; i++) {
-            // add reward to cumulativeAmount
+            uint256 cumulativeAmountIncrease = rewardPerPeriod * (i + 1);
```
folex, line 94: I'd name it without `increase`, as it misleads me into thinking that it is a per-period-delta. But it's the whole `cumulativeAmount` for the whole `rewardPeriod`.

PR #525, opened 2024-08-05
feat(kras): switch mainnet chain to raas

folex:


subgraph/fluence-graph.sh
```diff
@@ -123,60 +125,60 @@ done
...
+  SUBGRAPH_NETWORK="local"
+  ;;
+stage)
+  SUBGRAPH_NETWORK="stage"
+  ;;
+dar)
+  SUBGRAPH_NETWORK="dar"
+  ;;
+kras)
+  SUBGRAPH_NETWORK="mainnet"
```
justprosh, line 138: Can you write comment why it was renamed? Because on next lines we still have kras network

justprosh (approved):

subgraph/fluence-graph.sh
folex: Kras's The Graph now has two networks: `kras` as before, and `mainnet` pointing to raas chain.

